PR #1, opened 2021-06-03
Add some data for BlockPos

sciwhiz12:
Documents the copy constructor for `Matrix3f` and its parameter, documents the 3-int constructor and the codec and logger fields for `BlockPos`, and adds brief documentation for the `com.mojang.blaze3d` and `net.minecraft.client` packages.

### DIFF
--- a/data/classes/a.json
+++ b/data/classes/a.json
@@ -65,11 +65,7 @@
     },
     {
       "name": "<init>",
-      "descriptor": "(La;)V"
-    },
-    {
-      "name": "<init>",
-      "descriptor": "(Lb;)V",
+      "descriptor": "(La;)V",
       "javadoc": [
         "Copy constructor for this class."
       ],
@@ -80,6 +76,10 @@
           "javadoc": "The instance to copy from"
         }
       ]
+    },
+    {
+      "name": "<init>",
+      "descriptor": "(Lb;)V"
     },
     {
       "name": "<init>",
--- a/data/classes/a.json
+++ b/data/classes/a.json
@@ -69,7 +69,17 @@
     },
     {
       "name": "<init>",
-      "descriptor": "(Lb;)V"
+      "descriptor": "(Lb;)V",
+      "javadoc": [
+        "Copy constructor for this class."
+      ],
+      "parameters": [
+        {
+          "index": 1,
+          "name": "pOther",
+          "javadoc": "The instance to copy from"
+        }
+      ]
     },
     {
       "name": "<init>",
--- a/data/classes/fx.json
+++ b/data/classes/fx.json
@@ -1,9 +1,15 @@
 {
   "name": "fx",
+  "javadoc": [
+    "A position of a block in 3 dimensions."
+  ],
   "fields": [
     {
       "name": "a",
-      "descriptor": "Lcom/mojang/serialization/Codec;"
+      "descriptor": "Lcom/mojang/serialization/Codec;",
+      "javadoc": [
+        "The codec for a {@code BlockPos}."
+      ]
     },
     {
       "name": "b",
@@ -11,7 +17,10 @@
     },
     {
       "name": "e",
-      "descriptor": "Lorg/apache/logging/log4j/Logger;"
+      "descriptor": "Lorg/apache/logging/log4j/Logger;",
+      "javadoc": [
+        "The logger for the class."
+      ]
     },
     {
       "name": "f",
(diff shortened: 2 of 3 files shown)


data/classes/fx.json
```diff
@@ -57,7 +66,27 @@
     },
     {
       "name": "<init>",
-      "descriptor": "(III)V"
+      "descriptor": "(III)V",
+      "javadoc": [
+        "Constructs a {@code BlockPos} using the provided X, Y, and Z coordinates."
```
alcatrazEscapee, line 65: use the fully qualified class name as per standards ;)

sciwhiz12: Do we still need to use fully qualified class names, even in `{@code}` tags?

data/classes/a.json
```diff
@@ -69,7 +69,17 @@
     },
     {
       "name": "<init>",
-      "descriptor": "(Lb;)V"
+      "descriptor": "(Lb;)V",
+      "javadoc": [
+        "Copy constructor for this class."
+      ],
```
ChampionAsh5357, line 71: This is not the case. This takes a `Matrix4f` and converts it into a `Matrix3f`. This is used to get the resulting transformation or the normal matrix itself.

data/classes/fx.json
SizableShrimp: In code tags, they are not resolved, so no. However, why use a code tag instead of a link with a FQN?

data/classes/fx.json
```diff
@@ -1,17 +1,26 @@
...
+    "A position of a block in 3 dimensions."
+  ],
   "fields": [
     {
       "name": "a",
-      "descriptor": "Lcom/mojang/serialization/Codec;"
+      "descriptor": "Lcom/mojang/serialization/Codec;",
+      "javadoc": [
+        "The codec for a {@code BlockPos}."
+      ]
```
ChampionAsh5357: This is more of my opinion, but if a codec is javadoced, the format in which the data is preserved is useful to know. For example, the block pos serializes to/deserializes from an integer array of three elements.

data/classes/fx.json
sciwhiz12: A stylistic choice, since a `{@link}` tag to `BlockPos` would be redundant (as it would link to the same class it is already in). I've also seen javadocs in the JDK use plain text (`Constructs a BlockPos using the [...]`), but my preference is using the code tag for the monospace font.

data/classes/fx.json
```diff
@@ -1,17 +1,26 @@
...
       "name": "b",
       "descriptor": "Lfx;"
     },
     {
       "name": "e",
-      "descriptor": "Lorg/apache/logging/log4j/Logger;"
+      "descriptor": "Lorg/apache/logging/log4j/Logger;",
+      "javadoc": [
+        "The logger for the class."
+      ]
```
ChampionAsh5357: Comment is not useful. It's expected knowledge to know what a logger for a class is which violates the standards.

data/classes/fx.json
SizableShrimp: ah alright. seems good to me then

alcatrazEscapee: Fair nuff

ChampionAsh5357 (changes requested):

data/classes/fx.json
alcatrazEscapee: *Well akcuthally*, all you can specify is that it uses the `Codec.INT_STREAM`. Which invokes `DynamicOps#getIntStream`. Which itself is implementation specific. Yes, both `JsonOps` and `NBTDynamicOps` map this down to an int array when serialized to a data format. But codec's are data-format-agonstic. So this codec itself does not say it serializes to a int array *nessecarilly.*

ChampionAsh5357: If I wanted to correct my terminology, it would be more appropriate to say an int list of three elements. Either way, without the specific information on the expected format, this too would also violate simple/expected knowledge.

alcatrazEscapee: Except it doesn't require it being a list. You could, in theory, implement a dynamic ops that serialized streams as:
```json
{
    "stream this baby": true,
    "first": 0,
    "second": 2,
    "third": 2
}
```
You'd be quite stupid if you did, but you could. *I'll shutup now*

data/classes/a.json
sciwhiz12: Fixed to document the actual copy constructor.

data/classes/fx.json
sciwhiz12: For now, I've removed the javadoc for the codec field, since my change adds no real substantive information.

data/classes/fx.json
sciwhiz12: Removed javadoc to comply with standards.

data/classes/fx.json
```diff
@@ -57,7 +60,27 @@
     },
     {
       "name": "<init>",
-      "descriptor": "(III)V"
+      "descriptor": "(III)V",
+      "javadoc": [
+        "Constructs a {@code BlockPos} using the provided X, Y, and Z coordinates."
```
ChampionAsh5357, line 65: Currently waiting on Javadocs 1 before approving.

ChampionAsh5357 (approved):

SizableShrimp (approved):

PR #2, opened 2021-06-03
Add docs for standards

alcatrazEscapee:
Wrote the standards into two separate documents:

- [Rendered View](https://github.com/alcatrazEscapee/mappings-testing/blob/main/docs/standards.md) `standards.md` wraps the Parameter name / javadoc sections from the temp. working doc. Some minor wording was changed to target the document towards the audience of "people who want to submit mappings"
- [Rendered View](https://github.com/alcatrazEscapee/mappings-testing/blob/main/docs/review-merge-process.md) `review-merge-process.md` wraps the "Review / Merge process" from the temp working doc. (*real.. real shocker that one*). It's intended as a codification of the review and merge standards and procedures the team will adhere to and is worded as such.

I intentionally excluded the "Quality of Life tools" section as that's basically just brainstorming and dreams, and the CLA, because, well, it's already in the CLA bot.

### DIFF
--- a/docs/review-merge-process.md
+++ b/docs/review-merge-process.md
@@ -4,21 +4,21 @@ This document details the review and merge process for Parchment mappings.
 
 ### Requirements to merge:
 
-- If we have any automatic tests, they all must pass.
-  - Example: A PR may not be merged if the Contributor License Agreement (CLA) is not signed.
-- A PR requires two approving reviews
-  - One approving review is required to be from a Parchment team member.
-- We ask that people don’t review things that they don’t actually have the capability to effectively review (knowledge wise).
-- All external reviewers and suggestions are welcome.
+1. All reviews and suggestions from community members are welcome.
+1. All automatic tests must pass:
+    - The Contributor License Agreement (CLA) must be signed.
+1. A PR requires two approving reviews
+    - One approving review is required to be from a member of the ParchmentMC mapping reviewers team.
+    - The other review can be anyone except the original author.
+    - We ask that people don’t review things that they don’t actually have the capability to effectively review (knowledge wise).
+1. After a PR has been approved, it must wait at least 48 hours for any further comments (or, 24 hours if it is deemed to be only minor changes.)
+1. After the required waiting period, if there are no reviews that are requesting changes from a member of the mapping reviewers team then the PR will be merged.
+1. If, after such waiting period, there are members from the mapping reviewers team requesting changes, then those changes must be adressed before the PR can be merged, or go into the waiting period again.
+1. If a PR is generating conflicting opinions, it may be put to a fixed length comment period, where everone is open to disucss changes to the PR. After the comment period, lasting at most one week, it will be put to a vote between the ParchmentMC mapping reviewers team as to wether it should be merged or not.
+1. Submissions should be substantial, although are not required to have any 'minimum contribution' to be considered
+    - Prefer submitting one PR containing a number of smaller unrelated contributions, than many PRs containing very small contributions.
+    - On the other hand, do not lump together otherwise unrelated PR contributions unless there is a reason to do so (i.e. the unrelated contributions are small and can be easilly extended)
+1. PRs are merged and squashed, linked back to the author.
 
-On Resolving conflicting opinions:
 
-- Any names involving substantial conflicts should be available for public input for some length of time (i.e. a discussion time) before merging. After such time, the PR may be merged if there is:
-  1. A simple majority of Parchment team reviewers
-  2. In the event of a tie, A simple majority of all reviewers
-- Submissions should be substantial, although are not required to have any 'minimum contribution' to be considered
-  - Prefer submitting one PR containing a number of smaller unrelated contributions, than many PRs containing very small contributions.
-  - On the other hand, do not lump together otherwise unrelated PR contributions unless there is a reason to do so (i.e. the unrelated contributions are small and can be easilly extended)
-- Contributions must require a github authentication, to link contributors back to individual contributions.
-- Parchment Team / Reviewers can be free to merge and/or review at any time, based on time.
-- PRs are merged and squashed, linked back to the author.
+Note: the phrase "ParchmentMC mapping reviewers team" refers to the members of [the github team of the same name](https://github.com/orgs/ParchmentMC/teams/mapping-reviewers).
--- a/docs/review-merge-process.md
+++ b/docs/review-merge-process.md
@@ -9,15 +9,16 @@ This document details the review and merge process for Parchment mappings.
 - A PR requires two approving reviews
   - One approving review is required to be from a Parchment team member.
 - We ask that people don’t review things that they don’t actually have the capability to effectively review (knowledge wise).
-- All external reviewers are welcome, but not required.
+- All external reviewers and suggestions are welcome.
 
 On Resolving conflicting opinions:
 
 - Any names involving substantial conflicts should be available for public input for some length of time (i.e. a discussion time) before merging. After such time, the PR may be merged if there is:
   1. A simple majority of Parchment team reviewers
   2. In the event of a tie, A simple majority of all reviewers
-- Submissions should be substantial. (Single line typo fix PRs are discouraged)
-- Omnibus PRs may be made ocassionally to accumulate lots of smaller fixes.
+- Submissions should be substantial, although are not required to have any 'minimum contribution' to be considered
+  - Prefer submitting one PR containing a number of smaller unrelated contributions, than many PRs containing very small contributions.
+  - On the other hand, do not lump together otherwise unrelated PR contributions unless there is a reason to do so (i.e. the unrelated contributions are small and can be easilly extended)
 - Contributions must require a github authentication, to link contributors back to individual contributions.
 - Parchment Team / Reviewers can be free to merge and/or review at any time, based on time.
 - PRs are merged and squashed, linked back to the author.
--- a/docs/review-merge-process.md
+++ b/docs/review-merge-process.md
@@ -1,0 +1,23 @@
+# Mappings Review / Merge Process
+
+This document details the review and merge process for Parchment mappings.
+
+### Requirements to merge:
+
+- If we have any automatic tests, they all must pass.
+  - Example: A PR may not be merged if the Contributor License Agreement (CLA) is not signed.
+- A PR requires two approving reviews
+  - One approving review is required to be from a Parchment team member.
+- We ask that people don’t review things that they don’t actually have the capability to effectively review (knowledge wise).
+- All external reviewers are welcome, but not required.
+
+On Resolving conflicting opinions:
+
+- Any names involving substantial conflicts should be available for public input for some length of time (i.e. a discussion time) before merging. After such time, the PR may be merged if there is:
+  1. A simple majority of Parchment team reviewers
+  2. In the event of a tie, A simple majority of all reviewers
+- Submissions should be substantial. (Single line typo fix PRs are discouraged)
+- Omnibus PRs may be made ocassionally to accumulate lots of smaller fixes.
+- Contributions must require a github authentication, to link contributors back to individual contributions.
+- Parchment Team / Reviewers can be free to merge and/or review at any time, based on time.
+- PRs are merged and squashed, linked back to the author.
--- a/docs/standards.md
+++ b/docs/standards.md
@@ -1,0 +1,49 @@
+# Mapping Standards
+
+This document details the standards that all contributions to Parchment mappings must follow.
+
+### Parameter Names
+
+- Use American English for spellings
+  - Example: `color`, not `colour`, `armor`, not `armour`
+- All names should be prefixed with `p`
+  - Example: `pLevel`, `pXPos`
+- Names should make sense without the prefix, and should be valid java identifiers NOT reserved keywords as such:
+  - Example: `pFloat` is NOT valid, as `float` is a reserved keyword.
+  - Example:`pOs` is not a correct use of the prefix, it should be `pPos`
+- Parameter names should be based on current context but otherwise verbose and complete words
+  - Avoid single letters, or abbreviations
+    - Example: `pNorthWest` over `pNW`, `pMatrixStack` over `pMStack`. 
+  - Common abbreviations can be used: `IO` / `NBT`, etc.
+  - Use simple parameter names based off of class names, differentiating when there multiple of same type (using more words, not numbers)
+- Use Lower Camel Case (`lowerCamelCase`)
+- Do not use `$` or `_` in variable names
+- All parameters should match the regex `p[A-Z][A-Za-z0-9]+`.
+  - Example: `pDot`, `pPos`, `pBlockState`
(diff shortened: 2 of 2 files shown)


ChampionAsh5357:
I would make these ordered lists so we can refer to them on the fly (like if someone violates use american english names for spelling, I could say violates Params 1 instead of the actual requirement).

docs/standards.md
```diff
@@ -0,0 +1,52 @@
...
+2. All names should be prefixed with `p`
+    - Example: `pLevel`, `pXPos`
+3. Names should make sense without the prefix, and should be valid java identifiers NOT reserved keywords as such:
+    - Example: `pFloat` is NOT valid, as `float` is a reserved keyword.
+    - Example:`pOs` is not a correct use of the prefix, it should be `pPos`
+4. Parameter names should be based on current context but otherwise verbose and complete words
+5. Avoid single letters, or abbreviations.
+    - Example: `pNorthWest` over `pNW`, `pMatrixStack` over `pMStack`. 
+    - Exception: common abbreviations can be used: `IO` / `NBT`, etc.
+6. Use simple parameter names based off of class names, differentiating when there multiple of same type (using more words, not numbers)
```
ChampionAsh5357: Doesn't 4 and 6 contradict each other? Were they both in the doc and we just forgot removing one or the other?

alcatrazEscapee: They were both in the doc yes. I would agree they're not perfectly worded but disagree they're contradictory. i.e. `Level level` is both verbose (i.e. not `lvl`), a complete word, and based off the class names. Although I'm not sure how to reword these exactly to be more clear

ChampionAsh5357: Combine them into the same point then. `Parameter names should be based off of the class names and the current context. They should be verbose and valid words, differentiating when there are multiple of the same type.`

docs/standards.md
```diff
@@ -0,0 +1,52 @@
...
+    - Example: `pNorthWest` over `pNW`, `pMatrixStack` over `pMStack`. 
+    - Exception: common abbreviations can be used: `IO` / `NBT`, etc.
+6. Use Lower Camel Case (`lowerCamelCase`)
+7. Do not use `$` or `_` in variable names
+8. All parameters should match the regex `p[A-Z][A-Za-z0-9]+`.
+    - Example: `pDot`, `pPos`, `pBlockState`
+9. Favor using Mojang / Mojmap class names over <1.17 MCP classes. In general, the mojmap name is the de-facto standard (some exceptions may apply).
+    - Example: Use `level` over `world`, `blockEntity` over `tileEntity`
+10. Common parameter names can use abbreviations for brevity:
+    - Example: `BlockPos pos, Level level, BlockState state, WorldGenLevel level`
```
ChampionAsh5357: Just realized this parameter list itself is invalid if within a method signature. Want to avoid any ambiguity that might come or would result in a later PR of people wondering the same.

docs/standards.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Prioritize documenting public APIs over private fields and methods.
+- Prioritize methods and parameters, important fields, and classes.
+- Follow the principle of least effort for best payoff!
+- Do not document `package-info.java`.
+
+Content:
+
+1. Fully qualified mojmap class names should be used in javadoc tags such as `@link`.
+    - Example: `{@code net.minecraft.math.BlockPos}` instead of `{@code BlockPos`}
```
ChampionAsh5357: Based on https://github.com/ParchmentMC/mappings-testing/pull/1#discussion_r645109509, we should adjust the content information to include the same.

ChampionAsh5357 (changes requested):

ChampionAsh5357 (approved):

SizableShrimp (changes requested):
The content looks good to me. However, I might suggest some style changes to the general markdown formatting, e.g. using headers/labels over colons. Also moving `review-merge-process.md` to use an ordered list as well. Note that you can repeat `1.` over and over and still get them numbered. Might wanna do that to make adding new entries to the list easier later down the line, but personal preference. Otherwise seems good to merge.

docs/review-merge-process.md
```diff
@@ -0,0 +1,23 @@
+# Mappings Review / Merge Process
+
+This document details the review and merge process for Parchment mappings.
+
+### Requirements to merge:
+
+- If we have any automatic tests, they all must pass.
+  - Example: A PR may not be merged if the Contributor License Agreement (CLA) is not signed.
+- A PR requires two approving reviews
+  - One approving review is required to be from a Parchment team member.
```
sciwhiz12: This could be narrowed down to 'from a reviewer from the Parchment mapping reviewers team.'. Also, should this also note that reviews from the author do not count towards the two-review requirement?

docs/review-merge-process.md
```diff
@@ -0,0 +1,23 @@
...
+- We ask that people don’t review things that they don’t actually have the capability to effectively review (knowledge wise).
+- All external reviewers are welcome, but not required.
```
sciwhiz12: This sorta has a conflict with the `A PR requires two approving reviews; one must be from a Parchment team member` point: Who will the second review be from, if not from a Parchment team member? Do external reviews count towards the two-approvals requirement?

docs/review-merge-process.md
```diff
@@ -0,0 +1,23 @@
...
+
+On Resolving conflicting opinions:
+
+- Any names involving substantial conflicts should be available for public input for some length of time (i.e. a discussion time) before merging. After such time, the PR may be merged if there is:
+  1. A simple majority of Parchment team reviewers
+  2. In the event of a tie, A simple majority of all reviewers
```
sciwhiz12: What does "all reviewers" mean in this context? Because the first point specifies that a simple majority of all Parchment team reviewers, then it follows that the second point must be wider in scope than "all Parchment team reviewers". 
 - Does this mean that external (non mapping-team) reviewers count? 
 - Does this mean that all mapping team members (not just those in the reviewers subteam) are counted? 
 - What happens if we reasonably can't get the votes of a majority of the reviewers due to lack of activity on their part?

We should outline this process a bit more in a separate document (whether in this PR or in a future one).

docs/review-merge-process.md
```diff
@@ -0,0 +1,23 @@
...
+- Submissions should be substantial. (Single line typo fix PRs are discouraged)
+- Omnibus PRs may be made ocassionally to accumulate lots of smaller fixes.
+- Contributions must require a github authentication, to link contributors back to individual contributions.
+- Parchment Team / Reviewers can be free to merge and/or review at any time, based on time.
```
sciwhiz12: Should we have a small delay between the last required approval reivew and the actual merge? This would give both the reviewers team and external reviewers a window of time to confirm that nothing is majorly out of place.

docs/standards.md
```diff
@@ -0,0 +1,52 @@
...
+    - Example: `pFloat` is NOT valid, as `float` is a reserved keyword.
+    - Example: `pOs` is not a correct use of the prefix, it should be `pPos`
+4. Parameter names may be named based on the types, and the context in which they are used. They should be verbose and use complete words - do not omit essential information for brevity's sake.
+    - Example: `BlockPos pAdjacentPos, BlockPos pCurrentPos`, not `BlockPos pPos1, BlockPos pPos2`
+5. Avoid single letters, or abbreviations.
+    - Example: `pNorthWest` over `pNW`, `pMatrixStack` over `pMStack`. 
+    - Exception: common abbreviations can be used: `IO` / `NBT`, etc.
+6. Use Lower Camel Case (`lowerCamelCase`)
+7. Do not use `$` or `_` in variable names
+8. All parameters should match the regex `p[A-Z][A-Za-z0-9]+`.
```
sciwhiz12: Point 7 is also covered by Point 8 (as the regex does not include `$` and `_`).
Also, there should be a written explanation of the regex (in a subpoint), for people not fluent in regex.
> Parameters must start with the `p` prefix, followed by an uppercase letter, followed by one or more letters or numbers. Any other characters (such as `$` or `_` are not allowed).

docs/standards.md
```diff
@@ -0,0 +1,52 @@
...
+    - Example: `pDot`, `pPos`, `pBlockState`
+9. Favor using Mojang / Mojmap class names over <1.17 MCP classes. In general, the mojmap name is the de-facto standard (some exceptions may apply).
```
sciwhiz12: Is this a suggestion to use Mojmap/Official class names, or a restriction? Could we be given some examples of what exceptions may apply here (even as a comment on this PR)?

docs/standards.md
```diff
@@ -0,0 +1,52 @@
...
+    - Example: Use `level` over `world`, `blockEntity` over `tileEntity`
+10. Common parameter names can use abbreviations for brevity:
+    - Example: `BlockPos pos, BlockState state, WorldGenLevel level`
```
sciwhiz12: This should be moved as either a subpoint or the point directly after #<area>5. Also, the parameters should have the `p` prefix.

docs/standards.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Prioritize documenting public APIs over private fields and methods.
+- Prioritize methods and parameters, important fields, and classes.
+- Follow the principle of least effort for best payoff!
+- Do not document `package-info.java`.
+
+Content:
+
+1. Fully qualified mojmap class names should be used in javadoc tags such as `@link`.
+    - Example: `{@link net.minecraft.math.BlockPos}` instead of `{@link BlockPos`}
```
sciwhiz12:
```suggestion
    - Example: `{@link net.minecraft.math.BlockPos}` instead of `{@link BlockPos}`
```
Also, should this restriction only be in force for 'dumb' tags (tags which do not validate their contents), such as `{@code}`?

sciwhiz12 (changes requested):
I have a _few_ questions and changes to note. :grin:

docs/review-merge-process.md
alcatrazEscapee: > Who will the second review be from, if not from a Parchment team member?

Anyone who wants to review it.

> Do external reviews count towards the two-approvals requirement?

Yes, that is why two are required, but only one is mentioned as required to be from Parchment.

docs/review-merge-process.md
alcatrazEscapee: So based on my initial assumptions:

> Does this mean that external (non mapping-team) reviewers count?
> Does this mean that all mapping team members (not just those in the reviewers subteam) are counted?

I have not made a distinction between Parchment mapping reviewers and general members of Parchment - whichever team makes more sense to be used (probably mapping reviewers, which was not fully clarified at the time this document was written), makes sense here.

> What happens if we reasonably can't get the votes of a majority of the reviewers due to lack of activity on their part?

The the status quo (PR not merged) is maintained.

docs/review-merge-process.md
alcatrazEscapee: Unsure. A valid question to bring up. I don't terribly mind either way.

docs/standards.md
alcatrazEscapee: > Also, there should be a written explanation of the regex (in a subpoint), for people not fluent in regex.

This is *why* there are individual points explaining each note not covered by the regex (i.e. lower camel case, no special characters, p prefix.). Removing those would require one to understand this regex - hence why Point 7 exists.

docs/standards.md
alcatrazEscapee: Example: if the mojmap name has a typo in it, or is using non-american english spelling. (But yes, can add that)

docs/standards.md
alcatrazEscapee: There are two things here:

1. making the links work, in an IDE, with imports that may or may not exist
1. being able to simply / automatically migrate names across class/package changes between versions.

Point 1. would indicate only particular tags (`@link`, `@see`). Point 2 would suggest all. Thoughts?

docs/standards.md
sciwhiz12: Hmm, looks like I didn't read the list too closely. :eyes:

Point 6 should clarfied a bit because it may be confusing if `pLowerCamelCase` fits the lowerCamelCase requirement (as if we strip the prefix, it's now in PascalCase).
The other individual points should be moved closer to the regex (grouping the naming restrictions together), or a separate block after the standards list should be added to state that the regex is a consolidation of the given points.

docs/review-merge-process.md
sciwhiz12: Discussion on the Discord has led to a recommendataion of a minimum of 48 hours between the last approving review required for merging, and the actual merging itself.

However, I still firmly believe that an option of a minimum of 24 hours be inlcuded, for small uncontestable submissions (such as typo fixes, or minor grammatical changes).

docs/review-merge-process.md
sciwhiz12: For the first point, we should clarify that "Parchment team reviewers" means those part of the @ParchmentMC/mapping-reviewers subteam, which is expected to be those team members who wish to actively review PRs (and so hold the privilege of being able to merge those PRs in), while "all reviewers" means those part of the broader @ParchmentMC/mappings team, which holds all team members who manage the mappings data (but also includes those who aren't in the mapping-reviewers team, meaning those who are not able to review PRs due to time or personal constraints).

For the second point, since the mapping-reviewers subteam should only have those willing to be active in managing PRs and reviews, it should be fine. While the first vote will only be voted upon by the mapping-reviewers, those not voting can still express their own opinion, and those voting should take these statements into account in their votes.

docs/review-merge-process.md
sciwhiz12: Fair enough, though discussion in Discord has given thought to allowing small contribution PRs to only require one approving review instead of at least two.

docs/standards.md
sciwhiz12: While the idea of automatically migrating names across class/package changes is a good one, it would be more complex to implement because names may appear in a) `{@link}` and etc. tags (those which are checked by `javadoc` and the IDE), b) `{@code}` and etc. tags (those which do not have any checking), and c) without any tag (referencing names normally).

In my opinion, we should for now only require the use of fully qualified names for `{@link}` tags, and not 'dumb' tags or normal textual references to classes. In the future, we can reconsider this point once we start implementing automatic name migration.

docs/standards.md
```diff
@@ -0,0 +1,53 @@
...
+1. Avoid single letters, or abbreviations.
+    - Example: `pNorthWest` over `pNW`, `pMatrixStack` over `pMStack`. 
+    - Exception: common abbreviations can be used: `IO` / `NBT`, etc.
+    - Exception: common class / parameter names can be shortened: `BlockPos pPos, BlockState pState, WorldGenLevel pLevel`
+1. Use Lower Camel Case, treating the `p` prefix as the first word.
+    - Example: `pLowerCamelCase`
+3. Do not use `$` or `_` in variable names
+4. All parameters should match the regex `p[A-Z][A-Za-z0-9]*`.
+    - Simply, parameters should contain only alphanumeric characters, and start with the `p` prefix, followed by a capital letter.
+    - Example: `pDot`, `pPos`, `pBlockState`
```
sciwhiz12, line 25: Would've been good to also add an example of a name with a number character, but this is fine as-is.

sciwhiz12 (approved):
This looks good to me, with one small comment which you may or may not care about adding (fine by me either way).

SizableShrimp (approved):

PR #3, opened 2021-06-09
Add json schema files for the class jsons, package.json, and info.json

gigaherz:
Could have attempted to write regexp patterns for the version or descriptor strings, but that felt overkill and too easy to have mistakes in.


marchermans (approved):
It would be nice if you could generate one for the metadata as well?

gigaherz:
Will look into it.

sciwhiz12:
We may want to have this in ParchmentMC/Feather instead, since that's the library which has the objects and serializers for this format.

gigaherz:
I'm fine with that. My original thinking was to have this as close as possible to the data, so people who find the data can look at the schema and "know" how to interpret the data, without ever looking at the code in Feather, but I understand if that's not what the intention was.

Lanse505:
@sciwhiz12 @gigaherz 

Has this been moved and can be closed now?

gigaherz:
It has not, but it can be closed, I'll add the files to the PR at ParchmentMC/Feather#1

PR #9, opened 2021-06-21
Partially add parameter names to com/mojang/blaze3d and com/mojang/math

sciwhiz12:
Partially adds parameter names to classes within the `com/mojang/blaze3d` and `com/mojang/math` packages.

~~<sup>i was bored, okay?</sup>~~


data/com/mojang/blaze3d/audio/Library.mapping
```diff
@@ -1,15 +1,15 @@
 CLASS com/mojang/blaze3d/audio/Library
 	METHOD acquireChannel (Lcom/mojang/blaze3d/audio/Library$Pool;)Lcom/mojang/blaze3d/audio/Channel;
 		COMMENT Gets the source of a sound based on the mode
-		ARG 1 soundMode
+		ARG 1 pool
```
ChampionAsh5357: Comment needs to be adjusted to address changes.

data/com/mojang/blaze3d/audio/OggAudioStream.mapping
```diff
@@ -1,15 +1,17 @@
 CLASS com/mojang/blaze3d/audio/OggAudioStream
+	METHOD <init> (Ljava/io/InputStream;)V
+		ARG 1 input
 	METHOD convertMono (Ljava/nio/FloatBuffer;Lcom/mojang/blaze3d/audio/OggAudioStream$OutputConcat;)V
-		ARG 1 floatBuffer
-		ARG 2 oggAudioBuffer
+		ARG 1 buffer
```
ChampionAsh5357: Although a buffer, it is considered to be the 'left' or first channel according to the documentation on `STBVorbis#stb_vorbis_decode_frame_pushdata`. More appropriate to call it `channel` imo.

data/com/mojang/blaze3d/audio/OggAudioStream.mapping
```diff
@@ -1,15 +1,17 @@
...
+		ARG 2 output
 	METHOD convertStereo (Ljava/nio/FloatBuffer;Ljava/nio/FloatBuffer;Lcom/mojang/blaze3d/audio/OggAudioStream$OutputConcat;)V
-		ARG 1 soundChannel1
-		ARG 2 soundChannel2
-		ARG 3 oggAudioBuffer
+		ARG 1 leftBuffer
+		ARG 2 rightBuffer
```
ChampionAsh5357: `leftBuffer` -> `leftChannel`
`rightBuffer` -> `rightChannel`

data/com/mojang/blaze3d/audio/OggAudioStream.mapping
```diff
@@ -1,15 +1,17 @@
...
+		ARG 3 output
 	METHOD read (I)Ljava/nio/ByteBuffer;
 		ARG 1 size
 	METHOD readFrame (Lcom/mojang/blaze3d/audio/OggAudioStream$OutputConcat;)Z
-		ARG 1 oggAudioBuffer
+		ARG 1 output
 	CLASS OutputConcat
 		METHOD put (F)V
-			ARG 1 floatValue
+			ARG 1 value
```
ChampionAsh5357: According to `STBVorbis#stb_vorbis_decode_frame_pushdata`, this is the sample.

data/com/mojang/blaze3d/font/GlyphInfo.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS com/mojang/blaze3d/font/GlyphInfo
+	METHOD getAdvance (Z)F
+		ARG 1 useBoldOffset
```
ChampionAsh5357: `useBoldOffset` -> `isBold` or `bold`

Implementations suggest that this is only triggered when the font is bold or is hardcoded in.

data/com/mojang/blaze3d/font/TrueTypeGlyphProvider.mapping
```diff
@@ -1,7 +1,13 @@
 CLASS com/mojang/blaze3d/font/TrueTypeGlyphProvider
+	METHOD <init> (Ljava/nio/ByteBuffer;Lorg/lwjgl/stb/STBTTFontinfo;FFFFLjava/lang/String;)V
+		ARG 1 fontMemory
+		ARG 2 font
+		ARG 4 oversample
 	METHOD getGlyph (I)Lcom/mojang/blaze3d/font/TrueTypeGlyphProvider$Glyph;
 		ARG 1 character
 	CLASS Glyph
+		METHOD <init> (Lcom/mojang/blaze3d/font/TrueTypeGlyphProvider;IIIIFFI)V
+			ARG 8 index
```
ChampionAsh5357, line 10: I'm confused how this could be an argument of 8 when there's only seven parameters.

data/com/mojang/blaze3d/pipeline/RenderTarget.mapping
```diff
@@ -1,13 +1,16 @@
...
 		ARG 1 setViewport
 	METHOD _blitToScreen (IIZ)V
 		ARG 1 width
 		ARG 2 height
+		ARG 3 disableBlend
 	METHOD _resize (IIZ)V
 		ARG 1 width
 		ARG 2 height
-		ARG 3 onMac
+		ARG 3 checkError
```
ChampionAsh5357: `checkError` -> `clearError`

From `GL11#glGetError`:
> When GetError is called, the code is returned and the flag is cleared, so that a further error will again record its code

If we follow the method call, this leads us to `GlStateManager#_clear`. Since we do not get the error code back, I assume that this just clears the flag that notifies the user that an error is present.

data/com/mojang/blaze3d/pipeline/RenderTarget.mapping
```diff
@@ -16,20 +19,23 @@ CLASS com/mojang/blaze3d/pipeline/RenderTarget
 	METHOD blitToScreen (IIZ)V
 		ARG 1 width
 		ARG 2 height
+		ARG 3 disableBlend
 	METHOD clear (Z)V
-		ARG 1 onMac
+		ARG 1 checkError
```
ChampionAsh5357: `checkError` -> `clearError` as explained.

data/com/mojang/blaze3d/pipeline/RenderTarget.mapping
```diff
@@ -16,20 +19,23 @@ CLASS com/mojang/blaze3d/pipeline/RenderTarget
...
+	METHOD copyDepthFrom (Lcom/mojang/blaze3d/pipeline/RenderTarget;)V
+		ARG 1 otherTarget
 	METHOD createBuffers (IIZ)V
 		ARG 1 width
 		ARG 2 height
-		ARG 3 onMac
+		ARG 3 checkError
```
ChampionAsh5357: `checkError` -> `clearError` as explained.

data/com/mojang/blaze3d/pipeline/RenderTarget.mapping
```diff
@@ -16,20 +19,23 @@ CLASS com/mojang/blaze3d/pipeline/RenderTarget
...
 	METHOD resize (IIZ)V
 		ARG 1 width
 		ARG 2 height
-		ARG 3 onMac
+		ARG 3 checkError
```
ChampionAsh5357: `checkError` -> `clearError` as explained.

data/com/mojang/blaze3d/pipeline/TextureTarget.mapping
```diff
@@ -0,0 +1,6 @@
+CLASS com/mojang/blaze3d/pipeline/TextureTarget
+	METHOD <init> (IIZZ)V
+		ARG 1 width
+		ARG 2 height
+		ARG 3 useDepth
+		ARG 4 checkError
```
ChampionAsh5357: `checkError` -> `clearError` as explained.

data/com/mojang/blaze3d/platform/ClipboardManager.mapping
```diff
@@ -1,11 +1,11 @@
 CLASS com/mojang/blaze3d/platform/ClipboardManager
 	METHOD getClipboard (JLorg/lwjgl/glfw/GLFWErrorCallbackI;)Ljava/lang/String;
-		ARG 1 window
+		ARG 1 windowId
 		ARG 3 errorCallback
```
ChampionAsh5357, line 4: I don't know how there's a third argument when there are only two.

data/com/mojang/blaze3d/platform/ClipboardManager.mapping
```diff
@@ -1,11 +1,11 @@
 CLASS com/mojang/blaze3d/platform/ClipboardManager
 	METHOD getClipboard (JLorg/lwjgl/glfw/GLFWErrorCallbackI;)Ljava/lang/String;
-		ARG 1 window
+		ARG 1 windowId
```
ChampionAsh5357: Referred to as `window` in both Minecraft and GLFW.

data/com/mojang/blaze3d/platform/ClipboardManager.mapping
```diff
@@ -1,11 +1,11 @@
...
 		ARG 3 errorCallback
 	METHOD pushClipboard (JLjava/nio/ByteBuffer;[B)V
-		ARG 0 window
-		ARG 2 clipboardBuffer
+		ARG 0 windowId
```
ChampionAsh5357: `windowId` -> `window`.

See `GLFW#glfwSetClipboardString`.
Also highly unsure of numbering at this point.

data/com/mojang/blaze3d/platform/ClipboardManager.mapping
```diff
@@ -1,11 +1,11 @@
...
+		ARG 2 buffer
```
ChampionAsh5357, line 7: `buffer` -> `string` (assuming this is for the `ByteBuffer` param.

See `GLFW#glfwSetClipboardString`.

data/com/mojang/blaze3d/platform/ClipboardManager.mapping
```diff
@@ -1,11 +1,11 @@
...
 		ARG 3 clipboardContent
 	METHOD setClipboard (JLjava/lang/String;)V
-		ARG 1 window
-		ARG 3 string
+		ARG 1 windowId
```
ChampionAsh5357: `windowId` -> `window`.

Explained in other `windowId` calls.

data/com/mojang/blaze3d/platform/GlUtil.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS com/mojang/blaze3d/platform/GlUtil
+	METHOD allocateMemory (I)Ljava/nio/ByteBuffer;
+		ARG 0 size
+	METHOD freeMemory (Ljava/nio/Buffer;)V
+		ARG 0 buffer
```
ChampionAsh5357: `buffer` -> `ptr`

See `MemoryUtil#memFree`.

data/com/mojang/blaze3d/platform/GlUtil.mapping
```diff
@@ -0,0 +1,7 @@
...
+	METHOD populateSnooperWithOpenGL (Lcom/mojang/blaze3d/platform/SnooperAccess;)V
+		ARG 0 snooper
```
ChampionAsh5357: This is not the snooper itself but rather an access point to set certain fixed data. Might be more appropriate to fully qualify with `snooperAccess` or just `accessPoint`/`access`.

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
 		ARG 1 scanCode
 	METHOD getKey (Ljava/lang/String;)Lcom/mojang/blaze3d/platform/InputConstants$Key;
 		ARG 0 name
+	METHOD grabOrReleaseMouse (JIDD)V
+		ARG 0 windowId
```
ChampionAsh5357: `windowId` -> `window`

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
...
+		ARG 2 cursorMode
+		ARG 3 newX
```
ChampionAsh5357: `newX` -> `xPos`

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
...
+		ARG 5 newY
```
ChampionAsh5357: `newY` -> `yPos`

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
 		ARG 1 scanCode
 	METHOD getKey (Ljava/lang/String;)Lcom/mojang/blaze3d/platform/InputConstants$Key;
 		ARG 0 name
+	METHOD grabOrReleaseMouse (JIDD)V
+		ARG 0 windowId
+		ARG 2 cursorMode
```
ChampionAsh5357: `cursorMode` -> `cursorValue`.

See `GLFW#glfwSetInputMode`.

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
...
+		ARG 3 newX
+		ARG 5 newY
+	METHOD isKeyDown (JI)Z
+		ARG 0 windowId
```
ChampionAsh5357: `windowId` -> `window`

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
...
+		ARG 2 keyCode
```
ChampionAsh5357: `keyCode` -> `key`

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
...
+	METHOD setupKeyboardCallbacks (JLorg/lwjgl/glfw/GLFWKeyCallbackI;Lorg/lwjgl/glfw/GLFWCharModsCallbackI;)V
+		ARG 0 windowId
```
ChampionAsh5357: `windowId` -> `window`

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
...
+		ARG 2 keyCallback
+		ARG 3 charModifierCallback
+	METHOD setupMouseCallbacks (JLorg/lwjgl/glfw/GLFWCursorPosCallbackI;Lorg/lwjgl/glfw/GLFWMouseButtonCallbackI;Lorg/lwjgl/glfw/GLFWScrollCallbackI;Lorg/lwjgl/glfw/GLFWDropCallbackI;)V
+		ARG 0 windowId
```
ChampionAsh5357: `windowId` -> `window`

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
...
+		ARG 2 cursorPositionCallback
+		ARG 3 mouseButtonCallback
+		ARG 4 scrollCallback
+		ARG 5 dragAndDropCallback
+	METHOD updateRawMouseInput (JZ)V
+		ARG 0 windowId
```
ChampionAsh5357: `windowId` -> `window`

data/com/mojang/blaze3d/platform/InputConstants.mapping
```diff
@@ -4,7 +4,38 @@ CLASS com/mojang/blaze3d/platform/InputConstants
...
+		ARG 2 enableRawMouseInput
```
ChampionAsh5357: `enableRawMouseInput` -> `enableRawMouseMotion`

data/com/mojang/blaze3d/platform/Monitor.mapping
```diff
@@ -1,7 +1,9 @@
 CLASS com/mojang/blaze3d/platform/Monitor
+	METHOD <init> (J)V
+		ARG 1 monitorPointer
```
ChampionAsh5357: `monitorPointer` -> `monitor`

data/com/mojang/blaze3d/platform/MonitorCreator.mapping
```diff
@@ -0,0 +1,3 @@
+CLASS com/mojang/blaze3d/platform/MonitorCreator
+	METHOD createMonitor (J)Lcom/mojang/blaze3d/platform/Monitor;
+		ARG 1 monitorId
```
ChampionAsh5357: `monitorId` -> `monitor`

data/com/mojang/blaze3d/platform/NativeImage.mapping
```diff
@@ -11,13 +24,17 @@ CLASS com/mojang/blaze3d/platform/NativeImage
 		ARG 9 clamp
 		ARG 10 mipmap
 		ARG 11 autoClose
+	METHOD blendPixel (III)V
+		ARG 1 x
+		ARG 2 y
+		ARG 3 argb
```
ChampionAsh5357: This is the pixel color being blended into the specified pixel. So maybe `otherPixelColor` or something of the sort and then a comment specifying the format.

data/com/mojang/blaze3d/platform/NativeImage.mapping
```diff
@@ -11,13 +24,17 @@ CLASS com/mojang/blaze3d/platform/NativeImage
...
 	METHOD combine (IIII)I
 		ARG 0 alpha
 		ARG 1 blue
 		ARG 2 green
 		ARG 3 red
 	METHOD copyFrom (Lcom/mojang/blaze3d/platform/NativeImage;)V
-		ARG 1 from
+		ARG 1 otherImage
```
ChampionAsh5357: Need to standardize to either `other` or `other<Class>`.

data/com/mojang/blaze3d/platform/NativeImage.mapping
```diff
@@ -51,29 +68,41 @@ CLASS com/mojang/blaze3d/platform/NativeImage
 	METHOD fromBase64 (Ljava/lang/String;)Lcom/mojang/blaze3d/platform/NativeImage;
 		ARG 0 string
 	METHOD getA (I)I
-		ARG 0 col
+		ARG 0 packedARGB
```
ChampionAsh5357: Are 24-bit + alpha colors considered packed in the imaging world? If not, I would just leave this as `color` and leave a comment on the method specifying the bits used.

data/com/mojang/blaze3d/platform/NativeImage.mapping
```diff
@@ -51,29 +68,41 @@ CLASS com/mojang/blaze3d/platform/NativeImage
...
 	METHOD getB (I)I
-		ARG 0 col
+		ARG 0 packedARGB
```
ChampionAsh5357: Read 24-bit + alpha text.

data/com/mojang/blaze3d/platform/NativeImage.mapping
```diff
@@ -51,29 +68,41 @@ CLASS com/mojang/blaze3d/platform/NativeImage
...
+	METHOD getBlueOrLuminance (II)B
+		ARG 1 x
+		ARG 2 y
 	METHOD getG (I)I
-		ARG 0 col
+		ARG 0 packedARGB
```
ChampionAsh5357: Read 24-bit + alpha text.

data/com/mojang/blaze3d/platform/NativeImage.mapping
```diff
@@ -51,29 +68,41 @@ CLASS com/mojang/blaze3d/platform/NativeImage
...
+		ARG 2 y
 	METHOD getLuminanceOrAlpha (II)B
 		ARG 1 x
 		ARG 2 y
 	METHOD getPixelRGBA (II)I
 		ARG 1 x
 		ARG 2 y
 	METHOD getR (I)I
-		ARG 0 col
+		ARG 0 packedARGB
```
ChampionAsh5357: Read 24-bit + alpha text.

data/com/mojang/blaze3d/platform/NativeImage.mapping
```diff
@@ -83,10 +112,14 @@ CLASS com/mojang/blaze3d/platform/NativeImage
...
 		ARG 1 mipmap
+	METHOD setPixelLuminance (IIB)V
+		ARG 1 x
+		ARG 2 y
+		ARG 3 luminance
 	METHOD setPixelRGBA (III)V
 		ARG 1 x
 		ARG 2 y
-		ARG 3 value
+		ARG 3 rgba
```
ChampionAsh5357: `rgba` -> `color` and it's in ABGR format.

data/com/mojang/blaze3d/platform/NativeImage.mapping
```diff
@@ -118,8 +151,30 @@ CLASS com/mojang/blaze3d/platform/NativeImage
...
+			ARG 11 greenOffset
+			ARG 12 blueOffset
+			ARG 13 luminanceOffset
+			ARG 14 alphaOffset
+			ARG 15 supportedByStb
 		METHOD getStbFormat (I)Lcom/mojang/blaze3d/platform/NativeImage$Format;
 			ARG 0 channels
+	CLASS WriteCallback
+		METHOD <init> (Ljava/nio/channels/WritableByteChannel;)V
+			ARG 1 channel
```
ChampionAsh5357: `channel` -> `output`

data/com/mojang/blaze3d/platform/ScreenManager.mapping
```diff
@@ -1,8 +1,10 @@
 CLASS com/mojang/blaze3d/platform/ScreenManager
+	METHOD <init> (Lcom/mojang/blaze3d/platform/MonitorCreator;)V
+		ARG 1 creator
```
ChampionAsh5357: `creator` -> `monitorCreator`

data/com/mojang/blaze3d/platform/ScreenManager.mapping
```diff
@@ -1,8 +1,10 @@
...
 	METHOD clamp (III)I
-		ARG 0 minValue
-		ARG 1 value
-		ARG 2 maxValue
+		ARG 0 value
+		ARG 1 minimum
```
ChampionAsh5357: `minimum` -> `min`

data/com/mojang/blaze3d/platform/ScreenManager.mapping
```diff
@@ -1,8 +1,10 @@
...
+		ARG 2 maximum
```
ChampionAsh5357: `maximum` -> `max`

data/com/mojang/blaze3d/platform/SnooperAccess.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS com/mojang/blaze3d/platform/SnooperAccess
+	METHOD setFixedData (Ljava/lang/String;Ljava/lang/Object;)V
+		ARG 1 key
+		ARG 2 value
```
ChampionAsh5357: `value` -> `data`

data/com/mojang/blaze3d/platform/SnooperAccess.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS com/mojang/blaze3d/platform/SnooperAccess
+	METHOD setFixedData (Ljava/lang/String;Ljava/lang/Object;)V
+		ARG 1 key
```
ChampionAsh5357: `key` -> `id`

data/com/mojang/blaze3d/platform/TextureUtil.mapping
```diff
@@ -31,3 +31,6 @@ CLASS com/mojang/blaze3d/platform/TextureUtil
 		ARG 0 inputStream
 	METHOD releaseTextureId (I)V
 		ARG 0 textureId
+	METHOD writeAsPNG (Ljava/lang/String;IIII)V
+		ARG 0 filenamePrefix
```
ChampionAsh5357: I don't think prefix is what this is. This is the name of the texture itself. It's then suffixed with texture depth (assuming that's what the parameter is). So probably `textureName`.

data/com/mojang/blaze3d/platform/VideoMode.mapping
```diff
@@ -1,3 +1,16 @@
 CLASS com/mojang/blaze3d/platform/VideoMode
+	METHOD <init> (IIIIII)V
+		ARG 1 width
+		ARG 2 height
+		ARG 3 redBits
+		ARG 4 greenBits
+		ARG 5 blueBits
+		ARG 6 refreshRate
+	METHOD <init> (Lorg/lwjgl/glfw/GLFWVidMode$Buffer;)V
+		ARG 1 videoModeBuffer
```
ChampionAsh5357: `videoModeBuffer` -> `bufferVideoMode`. Just consistent naming scheme.

data/com/mojang/blaze3d/platform/Window.mapping
```diff
@@ -1,20 +1,27 @@
 CLASS com/mojang/blaze3d/platform/Window
+	METHOD <init> (Lcom/mojang/blaze3d/platform/WindowEventHandler;Lcom/mojang/blaze3d/platform/ScreenManager;Lcom/mojang/blaze3d/platform/DisplayData;Ljava/lang/String;Ljava/lang/String;)V
+		ARG 1 eventHandle
```
ChampionAsh5357: `eventHandler`

data/com/mojang/blaze3d/platform/Window.mapping
```diff
@@ -1,20 +1,27 @@
...
+		ARG 2 screenManager
+		ARG 3 displayData
+		ARG 4 videoMode
+		ARG 5 windowTitle
```
ChampionAsh5357: `windowTitle` -> `title`

data/com/mojang/blaze3d/platform/Window.mapping
```diff
@@ -24,33 +31,40 @@ CLASS com/mojang/blaze3d/platform/Window
 		ARG 4 framebufferHeight
 	METHOD onMove (JII)V
 		ARG 1 windowPointer
```
ChampionAsh5357: `windowPointer` -> `window`. Need to standardize naming set.

data/com/mojang/blaze3d/platform/Window.mapping
```diff
@@ -24,33 +31,40 @@ CLASS com/mojang/blaze3d/platform/Window
...
-		ARG 3 windowX
-		ARG 4 windowY
+		ARG 3 x
+		ARG 4 y
 	METHOD onResize (JII)V
 		ARG 1 windowPointer
```
ChampionAsh5357: `windowPointer` -> `window`

data/com/mojang/blaze3d/platform/Window.mapping
```diff
@@ -1,20 +1,27 @@
 CLASS com/mojang/blaze3d/platform/Window
+	METHOD <init> (Lcom/mojang/blaze3d/platform/WindowEventHandler;Lcom/mojang/blaze3d/platform/ScreenManager;Lcom/mojang/blaze3d/platform/DisplayData;Ljava/lang/String;Ljava/lang/String;)V
+		ARG 1 eventHandle
+		ARG 2 screenManager
+		ARG 3 displayData
+		ARG 4 videoMode
```
ChampionAsh5357: `videoMode` -> `preferredFullscreenVideoMode` since that is what the optional represents.

data/com/mojang/blaze3d/preprocessor/GlslPreprocessor.mapping
```diff
@@ -0,0 +1,19 @@
+CLASS com/mojang/blaze3d/preprocessor/GlslPreprocessor
+	METHOD isDirectiveDisabled (Ljava/lang/String;Ljava/util/regex/Matcher;I)Z
+		ARG 0 line
```
ChampionAsh5357: `line` -> `shaderData`. This is the full shader string which is then substringed in the method itself.

data/com/mojang/blaze3d/preprocessor/GlslPreprocessor.mapping
```diff
@@ -0,0 +1,19 @@
...
+		ARG 1 matcher
+	METHOD isDirectiveEnabled (Ljava/lang/String;Ljava/util/regex/Matcher;)Z
+		ARG 0 line
```
ChampionAsh5357: `line` -> `shaderData`

data/com/mojang/blaze3d/preprocessor/GlslPreprocessor.mapping
```diff
@@ -0,0 +1,19 @@
...
+		ARG 1 matcher
+	METHOD process (Ljava/lang/String;)Ljava/util/List;
+		ARG 1 line
```
ChampionAsh5357: `line` -> `shaderData`

data/com/mojang/blaze3d/preprocessor/GlslPreprocessor.mapping
```diff
@@ -0,0 +1,19 @@
...
+	METHOD processImports (Ljava/lang/String;Lcom/mojang/blaze3d/preprocessor/GlslPreprocessor$Context;Ljava/lang/String;)Ljava/util/List;
+		ARG 1 line
```
ChampionAsh5357: `line` -> `shaderData`

data/com/mojang/blaze3d/preprocessor/GlslPreprocessor.mapping
```diff
@@ -0,0 +1,19 @@
...
+		ARG 2 context
+		ARG 3 importPrefix
```
ChampionAsh5357: Not really the prefix, more of the initial directory to look for the imports within `./include`.

data/com/mojang/blaze3d/preprocessor/GlslPreprocessor.mapping
```diff
@@ -0,0 +1,19 @@
...
+	METHOD processVersions (Ljava/lang/String;Lcom/mojang/blaze3d/preprocessor/GlslPreprocessor$Context;)Ljava/lang/String;
+		ARG 1 line
```
ChampionAsh5357: `line` -> `versionData` as it just contains the initial line holding the version.

data/com/mojang/blaze3d/preprocessor/GlslPreprocessor.mapping
```diff
@@ -0,0 +1,19 @@
...
+		ARG 2 context
+	METHOD setVersion (Ljava/lang/String;I)Ljava/lang/String;
+		ARG 1 line
```
ChampionAsh5357: `line` -> `versionData`

data/com/mojang/blaze3d/shaders/EffectProgram.mapping
```diff
@@ -0,0 +1,11 @@
...
+	METHOD <init> (Lcom/mojang/blaze3d/shaders/Program$Type;ILjava/lang/String;)V
+		ARG 1 type
+		ARG 2 id
+		ARG 3 name
+	METHOD attachToEffect (Lcom/mojang/blaze3d/shaders/Effect;)V
+		ARG 1 effect
+	METHOD compileShader (Lcom/mojang/blaze3d/shaders/Program$Type;Ljava/lang/String;Ljava/io/InputStream;Ljava/lang/String;)Lcom/mojang/blaze3d/shaders/EffectProgram;
+		ARG 0 type
+		ARG 1 name
+		ARG 2 shaderStream
```
ChampionAsh5357: `shaderStream` -> `shaderData` since it's technically the resource data to be read.

data/com/mojang/blaze3d/shaders/Program.mapping
```diff
@@ -1,4 +1,26 @@
...
+	METHOD <init> (Lcom/mojang/blaze3d/shaders/Program$Type;ILjava/lang/String;)V
+		ARG 1 type
+		ARG 2 id
+		ARG 3 name
+	METHOD attachToShader (Lcom/mojang/blaze3d/shaders/Shader;)V
+		ARG 1 shader
+	METHOD compileShader (Lcom/mojang/blaze3d/shaders/Program$Type;Ljava/lang/String;Ljava/io/InputStream;Ljava/lang/String;Lcom/mojang/blaze3d/preprocessor/GlslPreprocessor;)Lcom/mojang/blaze3d/shaders/Program;
+		ARG 0 type
+		ARG 1 name
+		ARG 2 shaderStream
```
ChampionAsh5357: `shaderStream` -> `shaderData`

data/com/mojang/blaze3d/shaders/Program.mapping
```diff
@@ -1,4 +1,26 @@
...
+		ARG 4 preprocessor
+	METHOD compileShaderInternal (Lcom/mojang/blaze3d/shaders/Program$Type;Ljava/lang/String;Ljava/io/InputStream;Ljava/lang/String;Lcom/mojang/blaze3d/preprocessor/GlslPreprocessor;)I
+		ARG 0 type
+		ARG 1 name
+		ARG 2 shaderStream
```
ChampionAsh5357: `shaderStream` -> `shaderData`

data/com/mojang/blaze3d/shaders/ProgramManager.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS com/mojang/blaze3d/shaders/ProgramManager
+	METHOD glUseProgram (I)V
+		ARG 0 programId
```
ChampionAsh5357: `programId` -> `program`

data/com/mojang/blaze3d/systems/RenderSystem.mapping
```diff
@@ -0,0 +1,12 @@
+CLASS com/mojang/blaze3d/systems/RenderSystem
+	METHOD recordRenderCall (Lcom/mojang/blaze3d/pipeline/RenderCall;)V
+		ARG 0 renderCall
+	CLASS AutoStorageIndexBuffer
+		METHOD <init> (IILcom/mojang/blaze3d/systems/RenderSystem$AutoStorageIndexBuffer$IndexGenerator;)V
+			ARG 1 vertexStride
+			ARG 2 indexStride
+			ARG 3 generator
+		METHOD ensureStorage (I)V
+			ARG 1 newIndexCount
```
ChampionAsh5357: Interesting case on how to handle since technically this would be `indexCount` since you're setting the value but only if it is larger than the current value. Might be better to say `neededIndexCount` since we are ensuring that there is at least this much space.

data/com/mojang/blaze3d/vertex/BufferBuilder.mapping
```diff
@@ -15,10 +20,18 @@ CLASS com/mojang/blaze3d/vertex/BufferBuilder
 	METHOD putShort (IS)V
 		ARG 1 index
 		ARG 2 shortValue
+	METHOD putSortedQuadIndices (Lcom/mojang/blaze3d/vertex/VertexFormat$IndexType;)V
+		ARG 1 indexType
+	METHOD restoreSortState (Lcom/mojang/blaze3d/vertex/BufferBuilder$SortState;)V
+		ARG 1 state
```
ChampionAsh5357: `state` -> `sortState`

data/com/mojang/blaze3d/vertex/BufferBuilder.mapping
```diff
@@ -15,10 +20,18 @@ CLASS com/mojang/blaze3d/vertex/BufferBuilder
...
 	METHOD roundUp (I)I
 		ARG 0 x
+	METHOD setQuadSortOrigin (FFF)V
+		ARG 1 newSortX
+		ARG 2 newSortY
+		ARG 3 newSortZ
```
ChampionAsh5357: Would make sense to remove `new` since all this does is set the values.

data/com/mojang/blaze3d/vertex/BufferUploader.mapping
```diff
@@ -1,3 +1,13 @@
 CLASS com/mojang/blaze3d/vertex/BufferUploader
+	METHOD _end (Ljava/nio/ByteBuffer;Lcom/mojang/blaze3d/vertex/VertexFormat$Mode;Lcom/mojang/blaze3d/vertex/VertexFormat;ILcom/mojang/blaze3d/vertex/VertexFormat$IndexType;IZ)V
+		ARG 0 buffer
+		ARG 1 mode
+		ARG 2 format
+		ARG 3 vertices
```
ChampionAsh5357: `vertices` -> `vertexCount`

data/com/mojang/blaze3d/vertex/PoseStack.mapping
```diff
@@ -1,6 +1,8 @@
 CLASS com/mojang/blaze3d/vertex/PoseStack
 	METHOD mulPose (Lcom/mojang/math/Quaternion;)V
 		ARG 1 quaternion
+	METHOD mulPoseMatrix (Lcom/mojang/math/Matrix4f;)V
+		ARG 1 poseMatrix
```
ChampionAsh5357: `poseMatrix` -> `matrix`. Mojang also uses this method to just rotate a matrix within the debug renderer so we cannot declare it is specifically for the pos.

Most cases would use the pose from some other matrix however. So we could put `pose`, but this is a precedent matter.

data/com/mojang/blaze3d/vertex/VertexConsumer.mapping
```diff
@@ -24,15 +29,15 @@ CLASS com/mojang/blaze3d/vertex/VertexConsumer
 		ARG 1 u
 		ARG 2 v
 	METHOD putBulkData (Lcom/mojang/blaze3d/vertex/PoseStack$Pose;Lnet/minecraft/client/renderer/block/model/BakedQuad;FFFII)V
-		ARG 1 matrixEntry
+		ARG 1 poseEntry
```
ChampionAsh5357, line 32: Might want to document this as well for class names that are the same as field names as this does represent an entry in a `PoseStack`. Though, this doesn't have to be the case, it's assumed.

data/com/mojang/blaze3d/vertex/VertexMultiConsumer.mapping
```diff
@@ -1,8 +1,15 @@
 CLASS com/mojang/blaze3d/vertex/VertexMultiConsumer
+	METHOD create (Lcom/mojang/blaze3d/vertex/VertexConsumer;)Lcom/mojang/blaze3d/vertex/VertexConsumer;
+		ARG 0 consumer
 	METHOD create (Lcom/mojang/blaze3d/vertex/VertexConsumer;Lcom/mojang/blaze3d/vertex/VertexConsumer;)Lcom/mojang/blaze3d/vertex/VertexConsumer;
-		ARG 0 vertexBuilder
-		ARG 1 delegateBuilder
+		ARG 0 first
+		ARG 1 second
+	METHOD create ([Lcom/mojang/blaze3d/vertex/VertexConsumer;)Lcom/mojang/blaze3d/vertex/VertexConsumer;
+		ARG 0 consumers
```
ChampionAsh5357: `consumers` -> `delegates`

data/com/mojang/blaze3d/vertex/VertexMultiConsumer.mapping
```diff
@@ -1,8 +1,15 @@
 CLASS com/mojang/blaze3d/vertex/VertexMultiConsumer
+	METHOD create (Lcom/mojang/blaze3d/vertex/VertexConsumer;)Lcom/mojang/blaze3d/vertex/VertexConsumer;
+		ARG 0 consumer
```
ChampionAsh5357, line 3: I'm not sure what this would be since it refers to itself. Maybe `thisInstance`?

data/com/mojang/blaze3d/vertex/VertexMultiConsumer.mapping
```diff
@@ -40,3 +47,8 @@ CLASS com/mojang/blaze3d/vertex/VertexMultiConsumer
 			ARG 12 normalX
 			ARG 13 normalY
 			ARG 14 normalZ
+	CLASS Multiple
+		METHOD <init> ([Lcom/mojang/blaze3d/vertex/VertexConsumer;)V
+			ARG 1 delegates
+		METHOD forEach (Ljava/util/function/Consumer;)V
+			ARG 1 consumerConsumer
```
ChampionAsh5357, line 54: I hate this, but I have no idea what else to name it.

data/com/mojang/math/Matrix3f.mapping
```diff
@@ -3,5 +3,39 @@ CLASS com/mojang/math/Matrix3f
 		COMMENT Copy constructor for this class.
 		ARG 1 other
 			COMMENT The instance to copy from
+	METHOD <init> (Lcom/mojang/math/Matrix4f;)V
+		ARG 1 other
```
ChampionAsh5357: Not really the `other` matrix. It drops the 4th dimension of the 3x3, like becoming the normal or affine transform and throwing out the projected data in some higher plane dimension.

data/com/mojang/math/Quaternion.mapping
```diff
@@ -1,5 +1,42 @@
...
+		ARG 3 k
+		ARG 4 r
+	METHOD <init> (Lcom/mojang/math/Quaternion;)V
+		ARG 1 other
+	METHOD <init> (Lcom/mojang/math/Vector3f;FZ)V
+		ARG 1 vector
+		ARG 2 angle
+		ARG 3 degrees
+	METHOD cos (F)F
+		ARG 0 value
```
ChampionAsh5357: `value` -> `angle`.

See `Math#cos`.

data/com/mojang/math/Quaternion.mapping
```diff
@@ -1,5 +1,42 @@
...
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 other
+	METHOD fromXYZ (FFF)Lcom/mojang/math/Quaternion;
+		ARG 0 x
+		ARG 1 y
+		ARG 2 z
+	METHOD fromXYZ (Lcom/mojang/math/Vector3f;)Lcom/mojang/math/Quaternion;
+		ARG 0 vector
+	METHOD fromXYZDegrees (Lcom/mojang/math/Vector3f;)Lcom/mojang/math/Quaternion;
+		ARG 0 vectorDegrees
```
ChampionAsh5357: `vectorDegrees` -> `degreesVector`. Standardizing naming schemes.

data/com/mojang/math/Quaternion.mapping
```diff
@@ -1,5 +1,42 @@
...
+	METHOD cos (F)F
+		ARG 0 value
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 other
+	METHOD fromXYZ (FFF)Lcom/mojang/math/Quaternion;
+		ARG 0 x
+		ARG 1 y
+		ARG 2 z
+	METHOD fromXYZ (Lcom/mojang/math/Vector3f;)Lcom/mojang/math/Quaternion;
+		ARG 0 vector
```
ChampionAsh5357, line 22: `vector` -> `radiansVector`. Or maybe just leave a comment saying what kind of a vector it is (values held).

data/com/mojang/math/Quaternion.mapping
```diff
@@ -1,5 +1,42 @@
...
 	METHOD mul (Lcom/mojang/math/Quaternion;)V
-		ARG 1 quaternion
+		ARG 1 other
+	METHOD set (FFFF)V
+		ARG 1 i
+		ARG 2 j
+		ARG 3 k
+		ARG 4 r
+	METHOD sin (F)F
+		ARG 0 value
```
ChampionAsh5357: `value` -> `angle`

See `Math#sin`

data/com/mojang/math/SymmetricGroup3.mapping
```diff
@@ -0,0 +1,9 @@
+CLASS com/mojang/math/SymmetricGroup3
+	METHOD <init> (Ljava/lang/String;IIII)V
```
ChampionAsh5357, line 2: This method signature does not exist.

data/com/mojang/math/SymmetricGroup3.mapping
```diff
@@ -0,0 +1,9 @@
...
+		ARG 3 first
+		ARG 4 second
+		ARG 5 third
+	METHOD compose (Lcom/mojang/math/SymmetricGroup3;)Lcom/mojang/math/SymmetricGroup3;
+		ARG 1 other
+	METHOD permutation (I)I
+		ARG 1 index
```
ChampionAsh5357: Although correct, I think the mathematical term is element within the permutation set. Maybe a comment.

data/com/mojang/math/Vector3f.mapping
```diff
@@ -1,22 +1,37 @@
...
 		ARG 3 z
 	METHOD add (Lcom/mojang/math/Vector3f;)V
-		ARG 1 vector
+		ARG 1 other
 	METHOD clamp (FF)V
 		ARG 1 min
 		ARG 2 max
+	METHOD clamp (Lcom/mojang/math/Vector3f;Lcom/mojang/math/Vector3f;)V
+		ARG 1 lowerBound
+		ARG 2 higherBound
```
ChampionAsh5357: `min` and `max` instead.

data/com/mojang/math/Vector4f.mapping
```diff
@@ -1,6 +1,25 @@
 CLASS com/mojang/math/Vector4f
+	METHOD <init> (FFFF)V
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+		ARG 4 w
+	METHOD <init> (Lcom/mojang/math/Vector3f;)V
+		ARG 1 other
```
ChampionAsh5357: Not really `other` again. `vector3f` probably.

data/com/mojang/math/Vector4f.mapping
```diff
@@ -1,6 +1,25 @@
...
 	METHOD dot (Lcom/mojang/math/Vector4f;)F
-		ARG 1 vector
+		ARG 1 other
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 other
+	METHOD lerp (Lcom/mojang/math/Vector4f;F)V
+		ARG 1 other
+		ARG 2 percentage
+	METHOD mul (F)V
+		ARG 1 multiplier
```
ChampionAsh5357, line 22: I've seen `multiplier` and `scale`, needs standarization.

ChampionAsh5357 (changes requested):
Most of these name changes are based upon their implementation within LWJGL. We should try to stick to those names and then link to their appropriate javadoc via `@see`.

If we want to change any of those formats, we should have a discussion to standardize and then write it down somewhere.

data/com/mojang/blaze3d/font/TrueTypeGlyphProvider.mapping
sciwhiz12: I believe you have not counted the implicit first parameter for the instance itself. There's two implicit (not seen in source) parameters at the beginning: the instance of itself (`Glyph`) which is not shown in descriptors, and the instance of the owning outer class (`TrueTypeGlyphProvider`), which is shown in the descriptor. So that becomes parameter 0 (as parameter indexes start from 0) and parameter 1, respectively.

data/com/mojang/blaze3d/platform/ClipboardManager.mapping
sciwhiz12: `long`s and `double`s are counted twice for the parameter index, meaning that if a `long` parameter is at index 1, the next parameter index will be at 3.

See [JVMS 16, 4.3.3. "Method Descriptors"][jvms-4.3.3]: (emphasis mine)
> A method descriptor is valid only if it represents method parameters with a total length of 255 or less, where that length includes the contribution for `this` in the case of instance or interface method invocations. The total length is calculated by summing the contributions of the individual parameters, **where a parameter of type `long` or `double` contributes two units to the length and a parameter of any other type contributes one unit.**

[jvms-4.3.3]: https://docs.oracle.com/javase/specs/jvms/se16/html/jvms-4.html#jvms-4.3.3

data/com/mojang/blaze3d/platform/ClipboardManager.mapping
sciwhiz12: I would suggest against this change, as calling it `buffer` more accurately describes what the buffer is used for (maybe changing `scratchBuffer` could be considered). The actual string here is the third parameter (index 3), which is the actual string to be put into the clipboard contents. The buffer parameter is used by `setClipboard` to either use the scratch buffer if the string fits, or dynamically allocating a buffer if not.

data/com/mojang/blaze3d/platform/NativeImage.mapping
sciwhiz12: `otherPixelColor` implies that the color comes from another pixel, which while it may be true in nearly all cases, would add an unnecessary specificity as to the origin of the color, when `otherColor` or other names can suffice.

My opinion is that it be changed to `argbColor` (or perhaps renamed to `abgrColor`, as looking at the color unpacking helpers seems to imply that the actual order is `0xAABBGGRR`), which compromises between describing the value as a color and the format of the color.

data/com/mojang/blaze3d/platform/NativeImage.mapping
sciwhiz12: It is 'packed' in the sense that instead of 4 separates bytes for each channel (red, green, blue, alpha), they are instead packed into a single integer value. Also, instead of `color` I would suggest `abgr` or `abgrColor` (to be consistent with a comment from me above, I will use `abgrColor`).

data/com/mojang/blaze3d/platform/NativeImage.mapping
sciwhiz12: See https://github.com/ParchmentMC/Parchment/pull/9#discussion_r655807106.

data/com/mojang/blaze3d/platform/NativeImage.mapping
sciwhiz12: See https://github.com/ParchmentMC/Parchment/pull/9#discussion_r655807106.

data/com/mojang/blaze3d/platform/NativeImage.mapping
sciwhiz12: See https://github.com/ParchmentMC/Parchment/pull/9#discussion_r655807106.

data/com/mojang/blaze3d/platform/NativeImage.mapping
sciwhiz12: Renaming to `abgrColor`, see https://github.com/ParchmentMC/Parchment/pull/9#discussion_r655807106.

data/com/mojang/blaze3d/platform/ScreenManager.mapping
sciwhiz12: Would this be considered as an exception to the abbreviation rule (both `min` and `max)?

data/com/mojang/blaze3d/vertex/VertexMultiConsumer.mapping
sciwhiz12: I would suggest against that, and keep it at `consumer` because `thisInstance` looks to leaks an implementation detail (namely, that the instance is returned if there is only one consumer for delegation).

data/com/mojang/blaze3d/vertex/VertexMultiConsumer.mapping
sciwhiz12: :shipit:

data/com/mojang/math/Matrix3f.mapping
sciwhiz12: I feel like I got bitten by this same mistake in the past, because of the similarity of `Matrix3f` and `Matrix4f` in naming.

data/com/mojang/math/SymmetricGroup3.mapping
sciwhiz12: This is possibly an implication with enumerations specifically, where the string and integer represent the name and the ordinal of the enum value.

data/com/mojang/math/Vector4f.mapping
sciwhiz12: Looked at the other vector and matrix classes, and changed parameters so if the method name is `scale`, the parameter is `scale`, and if the method name is `mul`/`multiply`, then `multiplier`.

data/com/mojang/math/Quaternion.mapping
sciwhiz12: Added a comment instead, since the method names imply that unless specified, values are assumed to be in radians.

data/com/mojang/blaze3d/platform/ClipboardManager.mapping
ChampionAsh5357: I will say that `buffer` is fine then. The scratch buffer is only used if the data is smaller than 8kb. Otherwise, a buffer is delegated on the fly.

ChampionAsh5357 (dismissed):
I would say this is fine. Depending on the second reviewer's opinion, this will probably end up creating naming conventions that will need to be documented such that we remember to standardize.

marchermans (dismissed):

data/com/mojang/math/Vector4f.mapping
```diff
@@ -1,6 +1,25 @@
...
+		ARG 3 z
+		ARG 4 w
 	METHOD dot (Lcom/mojang/math/Vector4f;)F
-		ARG 1 vector
+		ARG 1 other
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 other
+	METHOD lerp (Lcom/mojang/math/Vector4f;F)V
+		ARG 1 other
+		ARG 2 percentage
```
alcatrazEscapee: `delta`, imo, following earlier conversations.

jriwanek (dismissed):

alcatrazEscapee (approved):

ChampionAsh5357 (approved):

PR #14, opened 2021-06-23
Random mappings

jaredlll08:


data/net/minecraft/client/Option.mapping
```diff
@@ -1,4 +1,6 @@
 CLASS net/minecraft/client/Option
+	METHOD <init> (Ljava/lang/String;)V
+		ARG 1 caption
```
ChampionAsh5357: `caption` -> `captionKey`. The caption itself is derived from this.

data/net/minecraft/client/Timer.mapping
```diff
@@ -1,3 +1,6 @@
 CLASS net/minecraft/client/Timer
+	METHOD <init> (FJ)V
+		ARG 1 tps
+		ARG 2 lastMS
```
ChampionAsh5357: `lastMS` -> `lastMs`

data/net/minecraft/client/Timer.mapping
```diff
@@ -1,3 +1,6 @@
 CLASS net/minecraft/client/Timer
+	METHOD <init> (FJ)V
+		ARG 1 tps
```
ChampionAsh5357: `tps` -> `ticksPerSec`. The param name is unclear and should be verbose.

data/net/minecraft/client/User.mapping
```diff
@@ -1,4 +1,11 @@
 CLASS net/minecraft/client/User
+	METHOD <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)V
+		ARG 1 name
+		ARG 2 uuid
+		ARG 3 accessToken
+		ARG 4 sessionType
```
ChampionAsh5357: `sessionType` -> `typeName`

data/net/minecraft/realms/RealmsScreen.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/realms/RealmsScreen
+	METHOD addLabel (Lnet/minecraft/realms/RealmsLabel;)Lnet/minecraft/realms/RealmsLabel;
+		ARG 1 label
+	METHOD row (I)I
+		ARG 0 offset
```
ChampionAsh5357: I'm not sure if `offset` is the best param name here. While yes, it offsets from that particular hardcoded position,  it's used more like a position. Offsets are usually more indicative of loops to which this method never seems to be used and is also supplied negative values at times.

data/net/minecraft/tags/TagCollection.mapping
```diff
@@ -7,10 +12,21 @@ CLASS net/minecraft/tags/TagCollection
 		ARG 1 resourceLocation
 	METHOD getTagOrEmpty (Lnet/minecraft/resources/ResourceLocation;)Lnet/minecraft/tags/Tag;
 		ARG 1 id
+	METHOD hasTag (Lnet/minecraft/resources/ResourceLocation;)Z
+		ARG 1 resourceLocation
```
ChampionAsh5357: `resourceLocation` -> `id`.

data/net/minecraft/tags/TagCollection.mapping
```diff
@@ -7,10 +12,21 @@ CLASS net/minecraft/tags/TagCollection
...
 	CLASS 1
 		METHOD getId (Lnet/minecraft/tags/Tag;)Lnet/minecraft/resources/ResourceLocation;
 			ARG 1 tag
 		METHOD getTagOrEmpty (Lnet/minecraft/resources/ResourceLocation;)Lnet/minecraft/tags/Tag;
 			ARG 1 id
+	CLASS NetworkPayload
+		METHOD <init> (Ljava/util/Map;)V
+			ARG 1 tags
+		METHOD read (Lnet/minecraft/network/FriendlyByteBuf;)Lnet/minecraft/tags/TagCollection$NetworkPayload;
+			ARG 0 byteBuf
```
ChampionAsh5357: Needs to be discussed on whether to have all buffers as `byteBuf` or `buffer`.

ChampionAsh5357 (changes requested):
Seems good, only thing that really needs discussion is `byteBuf` vs `buffer` depending on how Parchment wants to standardize these.

data/net/minecraft/client/Timer.mapping
eerussianguy: Not a big deal but TPS is a widely accepted term within the community for server performance

data/net/minecraft/client/User.mapping
jaredlll08: So the argument is passed into `Type#byName(String sessionType)`, which already had the `sessionType` name before I touched the class, would it not be inconsistent to use anything but `sessionType` since it is simply passed into that method?

data/net/minecraft/client/Timer.mapping
jaredlll08: > Not a big deal but TPS is a widely accepted term within the community for server performance


I wasn't sure how verbose we wanted to be, and ideally these mappings will be used by people just getting into modding the game and don't necessarily know the terms yet, so better to be more descriptive than less. I don't think the average vanilla player really knows what "tps" is (I don't even think there is a way to see tps on a vanilla client)

data/net/minecraft/tags/TagCollection.mapping
jaredlll08: Personally I like `buffer`, but when I was doing it I felt that using `byteBuf` would probably be safer and clearer as to what type of buffer

data/net/minecraft/client/User.mapping
ChampionAsh5357: Since `sessionType` is an MCP name, I would say just rename the parameter to be consistent. We're basing everything off the mojmap schema now so any MCP name that doesn't conform or just bad in general should be renamed.

For example, all `compound`s would be renamed to `tag` or whatever verbosity is needed.

ChampionAsh5357 (approved):

sciwhiz12 (approved):

Darkhax (approved):

PR #35, opened 2021-06-30
mapped part of net.minecraft.core

justliliandev:
tbh, I don't know most of the other net.minecraft.core stuff


data/net/minecraft/core/BlockPos.mapping
```diff
@@ -86,6 +94,7 @@ CLASS net/minecraft/core/BlockPos
 		ARG 0 pos
 		ARG 2 direction
 	METHOD randomBetweenClosed (Ljava/util/Random;IIIIIII)Ljava/lang/Iterable;
+		COMMENT Creates an Iterable with n Objects in the area.
```
jaredlll08: This comment doesn't really make sense. Firstly `n` is a bad place holder here, I know it was used in the data we worked off, but back then the param was also called `n`.

Also the capitalisation of `Objects` just adds more confusion.

data/net/minecraft/core/BlockPos.mapping
```diff
@@ -94,12 +103,24 @@ CLASS net/minecraft/core/BlockPos
 		ARG 5 maxX
 		ARG 6 maxY
 		ARG 7 maxZ
+	METHOD randomInCube (Ljava/util/Random;ILnet/minecraft/core/BlockPos;I)Ljava/lang/Iterable;
+		COMMENT Creates an Iterable with n BlockPos Objects in a cube around the center with the specified maxDistance.
```
jaredlll08: Once again, the comment doesn't make sense, no need to specify `BlockPos Objects` and I would argue that saying it makes an iterable is pointless.

Honestly what is here is basically just the `@return` javadoc imo.

data/net/minecraft/core/BlockPos.mapping
```diff
@@ -94,12 +103,24 @@ CLASS net/minecraft/core/BlockPos
...
 	METHOD relative (Lnet/minecraft/core/Direction;I)Lnet/minecraft/core/BlockPos;
 		COMMENT Offsets this BlockPos n blocks in the given direction
 		ARG 1 facing
 		ARG 2 n
 	METHOD rotate (Lnet/minecraft/world/level/block/Rotation;)Lnet/minecraft/core/BlockPos;
 		ARG 1 rotation
+	METHOD south (I)Lnet/minecraft/core/BlockPos;
+		ARG 1 distance
+	METHOD subtract (Lnet/minecraft/core/Vec3i;)Lnet/minecraft/core/BlockPos;
+		ARG 1 vec3i
```
jaredlll08: This should be `vector` to match with the param name in `offset`

data/net/minecraft/core/Cursor3D.mapping
```diff
@@ -1 +1,15 @@
-CLASS net/minecraft/core/Cursor3D
+CLASS net/minecraft/core/Cursor3D net/minecraft/core/Cursor3D
+	COMMENT This class represents a cuboid. You can iterate over it using the methods {@link Cursor3D#advance()} and {@link Cursor3D#nextX()}, {@link Cursor3D#nextY()}, {@link Cursor3D#nextZ()}
```
jaredlll08: bit nitpicky, but I think the correct english is:
```
	COMMENT This class represents a cuboid. You can iterate over it using the methods {@link Cursor3D#advance()}, {@link Cursor3D#nextX()}, {@link Cursor3D#nextY()} and {@link Cursor3D#nextZ()}
```

data/net/minecraft/core/Cursor3D.mapping
```diff
@@ -1 +1,15 @@
...
+	METHOD <init> (IIIIII)V
+		ARG 1 originX
+		ARG 2 originY
+		ARG 3 originZ
+		ARG 4 x2
+		ARG 5 y2
+		ARG 6 z2
```
jaredlll08: I was mapping this class at some point before I had to go do other things, I think the best names for these would be 
```
destinationX,
destinationY
destinationZ
```


Simply to match with the `originX` , `originY` and `originZ` that is already there.

data/net/minecraft/core/Cursor3D.mapping
```diff
@@ -1 +1,15 @@
...
+	METHOD getNextType ()I
+		COMMENT Returns the Type of the next BlockPos.
```
jaredlll08: Few things...
1) why is `Type` in caps, it isn't a class.
2) What does BlockPos have to do with this? 
The whole class works with ints and is keeping track of the individual int, yes they are being put into a BlockPos later down the line, but you are not getting a "type" of a BlockPos or even strictly working with BlockPos at all.
3) This comment is incredibly misleading, BlockPos don't have a Type.

data/net/minecraft/core/Direction.mapping
```diff
@@ -9,6 +9,14 @@ CLASS net/minecraft/core/Direction
 		COMMENT Normalized Vector that points in the direction of this Facing
 	FIELD oppositeIndex I
 		COMMENT Index of the opposite Facing in the VALUES array
+	METHOD <init> (Ljava/lang/String;IIIILjava/lang/String;Lnet/minecraft/core/Direction$AxisDirection;Lnet/minecraft/core/Direction$Axis;Lnet/minecraft/core/Vec3i;)V
+		ARG 3 index
+		ARG 4 oppositeIndex
+		ARG 5 horizontalIndex
```
jaredlll08: So for `ARG 3` and `ARG 5`, they are just being set to field names, `data3d` and `data2d` respectively, so naming them differently could be confusing for users.

I did discuss it in a private channel, but here is a repost of my message:
```java
byte b0, b1;
public thing(byte pitch, byte yaw){
  this.b0 = pitch;
  this.b1 = yar;
}

public byte getB0() {
  return b0;
}
public byte getB1() {
  return b1;
}
```

Sure the values may actually be `pitch` and `yaw`, but if someone was using it, their IDE would show them this, and be pretty confusing as to what is what.
image

I think doing `@param data3d <explanation here>` would be the best way to go about explaining what it does.

data/net/minecraft/core/IdMapper.mapping
```diff
@@ -1,11 +1,15 @@
 CLASS net/minecraft/core/IdMapper
+	METHOD <init> (I)V
+		ARG 1 expectedMaxSize
```
jaredlll08: I think this is maybe a bit *too* verbose?
Just having
```
maxSize
```
should be fine

data/net/minecraft/core/IdMapper.mapping
```diff
@@ -1,11 +1,15 @@
...
 	METHOD add (Ljava/lang/Object;)V
 		ARG 1 key
 	METHOD addMapping (Ljava/lang/Object;I)V
-		ARG 1 key
-		ARG 2 value
+		ARG 1 value
+		ARG 2 id
```
jaredlll08: ARG 1 is a key, not a value.

data/net/minecraft/core/Rotations.mapping
```diff
@@ -5,6 +5,14 @@ CLASS net/minecraft/core/Rotations
 		COMMENT Rotation on the Y axis
 	FIELD z F
 		COMMENT Rotation on the Z axis
+	METHOD <init> (FFF)V
+		ARG 1 x
+		ARG 2 y
+		ARG 3 z
+	METHOD <init> (Lnet/minecraft/nbt/ListTag;)V
+		ARG 1 tag
```
jaredlll08: I think this should be a bit more verbose, since this specifically needs to be a `ListTag`.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -10,14 +11,14 @@ CLASS net/minecraft/core/Vec3i
 		ARG 2 y
 		ARG 3 z
 	METHOD above ()Lnet/minecraft/core/Vec3i;
-		COMMENT Offset this vector 1 block up
+		COMMENT Creates a new Vec3i one unit up.
 	METHOD above (I)Lnet/minecraft/core/Vec3i;
-		COMMENT Offset this vector upwards by the given distance.
+		COMMENT Creates a new Vec3i up by the distance.
```
jaredlll08: This is technically incorrect, since if the distance is `0`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -10,14 +11,14 @@ CLASS net/minecraft/core/Vec3i
...
 		ARG 1 distance
 	METHOD below ()Lnet/minecraft/core/Vec3i;
-		COMMENT Offset this vector 1 block down
+		COMMENT Creates a new Vec3i one unit down.
 	METHOD below (I)Lnet/minecraft/core/Vec3i;
-		COMMENT Offset this vector downwards by the given distance.
+		COMMENT Creates a new Vec3i down by the distance.
```
jaredlll08: This is technically incorrect, since if the distance is `0`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -39,12 +40,15 @@ CLASS net/minecraft/core/Vec3i
...
+		COMMENT Calculate squared distance to the given Vector.
+		ARG 1 to
 	METHOD distSqr (Lnet/minecraft/core/Vec3i;Z)D
-		ARG 1 vector
+		ARG 1 to
 		ARG 2 useCenter
+	METHOD east ()Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i one unit east.
 	METHOD east (I)Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i east by the distance.
```
jaredlll08: This is technically incorrect, since if the distance is `0`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -56,27 +60,42 @@ CLASS net/minecraft/core/Vec3i
...
 		COMMENT Gets the Z coordinate.
+	METHOD lambda$static$0 ([I)Lnet/minecraft/core/Vec3i;
+		ARG 0 data
+	METHOD lambda$static$1 (Ljava/util/stream/IntStream;)Lcom/mojang/serialization/DataResult;
+		ARG 0 intStream
+	METHOD lambda$static$2 (Lnet/minecraft/core/Vec3i;)Ljava/util/stream/IntStream;
+		ARG 0 vec3i
 	METHOD multiply (I)Lnet/minecraft/core/Vec3i;
-		ARG 1 scalar
+		COMMENT Creates a new Vec3i with all values multiplied with the factor.
```
jaredlll08: This is technically incorrect, since if the distance is `1`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -56,27 +60,42 @@ CLASS net/minecraft/core/Vec3i
...
+		ARG 1 factor
```
jaredlll08: We discussed this in my PR. Scalar is the correct term for this.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -56,27 +60,42 @@ CLASS net/minecraft/core/Vec3i
...
+	METHOD north ()Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i one unit north.
 	METHOD north (I)Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i north by the distance.
```
jaredlll08: This is technically incorrect, since if the distance is `0`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -56,27 +60,42 @@ CLASS net/minecraft/core/Vec3i
...
 		ARG 1 distance
 	METHOD offset (DDD)Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i with the given offset. The values are getting floored.
```
jaredlll08: This is technically incorrect, since if `x, y, z` are `0`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -56,27 +60,42 @@ CLASS net/minecraft/core/Vec3i
...
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
 	METHOD offset (III)Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i with the given offset.
```
jaredlll08: This is technically incorrect, since if `x, y, z` are `0`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -56,27 +60,42 @@ CLASS net/minecraft/core/Vec3i
...
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
 	METHOD offset (Lnet/minecraft/core/Vec3i;)Lnet/minecraft/core/Vec3i;
-		ARG 1 vector
+		COMMENT Creates a new Vec3i with the given offset.
```
jaredlll08: This is technically incorrect, since if the given vector is (0,0,0) then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -56,27 +60,42 @@ CLASS net/minecraft/core/Vec3i
...
+		ARG 1 offset
 	METHOD relative (Lnet/minecraft/core/Direction$Axis;I)Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i with the distance on the axis.
```
jaredlll08: This is technically incorrect, since if the distance is `0`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -56,27 +60,42 @@ CLASS net/minecraft/core/Vec3i
...
 		ARG 1 axis
 		ARG 2 distance
 	METHOD relative (Lnet/minecraft/core/Direction;)Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i offseted by 1 in the direction.
 		ARG 1 direction
 	METHOD relative (Lnet/minecraft/core/Direction;I)Lnet/minecraft/core/Vec3i;
-		COMMENT Offsets this Vector by the given distance in the specified direction.
+		COMMENT Creates a new Vec3i with the distance in the direction.
```
jaredlll08: This is technically incorrect, since if the distance is `0`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -85,9 +104,16 @@ CLASS net/minecraft/core/Vec3i
 		ARG 1 y
 	METHOD setZ (I)Lnet/minecraft/core/Vec3i;
 		ARG 1 z
+	METHOD south ()Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i one unit south.
 	METHOD south (I)Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i south by the distance.
```
jaredlll08: This is technically incorrect, since if the distance is `0`, then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -85,9 +104,16 @@ CLASS net/minecraft/core/Vec3i
...
 		ARG 1 distance
 	METHOD subtract (Lnet/minecraft/core/Vec3i;)Lnet/minecraft/core/Vec3i;
-		ARG 1 vector
+		COMMENT Creates a new Vec3i subtracted by the subtrahend.
```
jaredlll08: This is technically incorrect, since if the vector is `(0, 0, 0)` then it returns the same vector object.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -85,9 +104,16 @@ CLASS net/minecraft/core/Vec3i
...
+		ARG 1 subtrahend
```
jaredlll08: I'm really not a fan of this name, it *may* be technically correct (this is the first time I've ever heard the word and googling it says `quantity or number`, which I don't think a Vector specifically is (although I could be wrong.)

Either way, if someone has to google what it means, I don't think it is a good name.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -85,9 +104,16 @@ CLASS net/minecraft/core/Vec3i
...
+	METHOD west ()Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i one unit west.
 	METHOD west (I)Lnet/minecraft/core/Vec3i;
+		COMMENT Creates a new Vec3i west by the distance.
```
jaredlll08: This is technically incorrect, since if the distance is `0`, then it returns the same vector object.

jaredlll08 (changes requested):

data/net/minecraft/core/Rotations.mapping
Lanse505: Seconded by Me and ChampionAsh

data/net/minecraft/core/Vec3i.mapping
justliliandev: Any idea on how to phrase it, so it's clear it creates a new Vec3i if the distance is not 0, without being too long?

jaredlll08:
```
If distance is not 0, will create a new Vec3i offset by the distance upwards, otherwise will return this vector.
```
something like that maybe?

data/net/minecraft/core/BlockPos.mapping
```diff
@@ -9,6 +9,12 @@ CLASS net/minecraft/core/BlockPos
 			COMMENT The Y coordinate
 		ARG 3 z
 			COMMENT The Z coordinate
+	METHOD <init> (Lnet/minecraft/core/Position;)V
+		ARG 1 position
+	METHOD <init> (Lnet/minecraft/core/Vec3i;)V
+		ARG 1 vec3i
```
alcatrazEscapee, line 15: `vector`

data/net/minecraft/core/BlockPos.mapping
```diff
@@ -9,6 +9,12 @@ CLASS net/minecraft/core/BlockPos
...
+	METHOD <init> (Lnet/minecraft/world/phys/Vec3;)V
+		ARG 1 vec3
```
alcatrazEscapee, line 17: `vector`

data/net/minecraft/core/BlockPos.mapping
```diff
@@ -56,7 +64,7 @@ CLASS net/minecraft/core/BlockPos
 		ARG 0 aabb
 	METHOD cross (Lnet/minecraft/core/Vec3i;)Lnet/minecraft/core/BlockPos;
 		COMMENT Calculate the cross product of this and the given Vector
-		ARG 1 vec
+		ARG 1 crossWith
```
alcatrazEscapee, line 67: `vector` or `other`. `crossWith` is redundant as the method is already named `cross`. The comment above, either "Vector" is the name of a class and should be `{@code Vector}`, or it's the textual name and should not be capitalized.

data/net/minecraft/core/BlockPos.mapping
```diff
@@ -86,6 +94,7 @@ CLASS net/minecraft/core/BlockPos
 		ARG 0 pos
 		ARG 2 direction
 	METHOD randomBetweenClosed (Ljava/util/Random;IIIIIII)Ljava/lang/Iterable;
+		COMMENT @return an Iterable of blockPositions with the given amount of elements in the area.
```
alcatrazEscapee, line 97: This comment adds nothing useful.

- There's no need to document the return type, that's clear from the method itself
- `Iterable`, if a class name, should be in `{@code }`, otherwise it's inconsistent grammar.
- `blockPositions` is not a word nor is it proper english
- "with the given amount of elements in the area" means absolutely nothing.

```java
/**
 * Iterates through {@code amount} randomly chosen positions within a cube, defined by the provided minimum and maximum positions.
 */
```

data/net/minecraft/core/BlockPos.mapping
```diff
@@ -94,12 +103,24 @@ CLASS net/minecraft/core/BlockPos
 		ARG 5 maxX
 		ARG 6 maxY
 		ARG 7 maxZ
+	METHOD randomInCube (Ljava/util/Random;ILnet/minecraft/core/BlockPos;I)Ljava/lang/Iterable;
+		COMMENT @return an Iterable of blockPositions with the given amount of elements in a cube around the center with the given maxDistance.
```
alcatrazEscapee, line 107: Same here, many issues with this explanation. It doesn't say what it *does*, it's just badly worded. For example:
```java
/**
 * Iterates through {@code amount} randomly chosen positions within a cube, defined by a {@code center} and a maximum manhattan {@code distance} away.
 */
```

data/net/minecraft/core/Direction.mapping
```diff
@@ -65,6 +84,21 @@ CLASS net/minecraft/core/Direction
...
+		ARG 0 direction
+	METHOD lambda$static$4 (Lnet/minecraft/core/Direction;)I
+		ARG 0 direction
+	METHOD lambda$static$6 (Lnet/minecraft/core/Direction;)Ljava/lang/Long;
+		ARG 0 direction
+	METHOD lambda$static$7 (Lnet/minecraft/core/Direction;)Lnet/minecraft/core/Direction;
+		ARG 0 direction
+	METHOD lambda$static$8 (Lnet/minecraft/core/Direction;Lnet/minecraft/core/Direction;)Lnet/minecraft/core/Direction;
+		ARG 0 direction1
+		ARG 1 direction2
```
alcatrazEscapee, line 101: These are very meaningless names, is there a reason why they can't be named properly?

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -10,14 +11,14 @@ CLASS net/minecraft/core/Vec3i
 		ARG 2 y
 		ARG 3 z
 	METHOD above ()Lnet/minecraft/core/Vec3i;
-		COMMENT Offset this vector 1 block up
+		COMMENT Creates a new Vec3i one unit up.
```
alcatrazEscapee, line 14: Again, class names, i.e. code, should be in an `{@code }` tag.

But moreover, these docs are... bordering on completely redundant. They don't particularly add anything that isn't immediately obvious from looking at the method. I'd almost be inclined to say the original javadocs here were better as these edits are pedantic and basically describing the code exactly as is.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -26,10 +27,10 @@ CLASS net/minecraft/core/Vec3i
 		ARG 1 vector
 		ARG 2 distance
 	METHOD cross (Lnet/minecraft/core/Vec3i;)Lnet/minecraft/core/Vec3i;
-		COMMENT Calculate the cross product of this and the given Vector
-		ARG 1 vector
+		COMMENT Calculate the cross product of this and the given Vector.
+		ARG 1 crossWith
```
alcatrazEscapee, line 31: Again, `crossWith` I think is redundant.

data/net/minecraft/core/Vec3i.mapping
```diff
@@ -39,44 +40,65 @@ CLASS net/minecraft/core/Vec3i
...
+	METHOD lambda$static$0 ([I)Lnet/minecraft/core/Vec3i;
+		ARG 0 data
+	METHOD lambda$static$1 (Ljava/util/stream/IntStream;)Lcom/mojang/serialization/DataResult;
+		ARG 0 intStream
+	METHOD lambda$static$2 (Lnet/minecraft/core/Vec3i;)Ljava/util/stream/IntStream;
+		ARG 0 vec3i
 	METHOD multiply (I)Lnet/minecraft/core/Vec3i;
+		COMMENT If scalar is 1 will return this vector.
+		COMMENT If scalar is 0 will return {@link Vec3i.ZERO}.
+		COMMENT Will create a new Vec3i with all values multiplied by the scalar.
```
alcatrazEscapee, line 72: Again, writing down every possible implementation detail of the method isn't really important. Vec3i is immutable to my knowledge, meaning that all the caller needs to know is this method multiplies all coordinate values by a scalar value and returns the result. Not which specific instance will get returned (as really, that's an memory saving optimization and not at all relevant to the javadoc)

data/net/minecraft/core/cauldron/CauldronInteraction.mapping
```diff
@@ -0,0 +1,128 @@
+CLASS net/minecraft/core/cauldron/CauldronInteraction net/minecraft/core/cauldron/CauldronInteraction
+	COMMENT The maps in this class are for each vanilla cauldron block. The EMPTY map handles everything, that can be done with an empty cauldron, the map itself is not empty.
+	METHOD addDefaultInteractions (Ljava/util/Map;)V
+		ARG 0 cauldronInteractionMap
+	METHOD emptyBucket (Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/sounds/SoundEvent;)Lnet/minecraft/world/InteractionResult;
+		ARG 0 level
+		ARG 1 pos
+		ARG 2 player
+		ARG 3 interactionHand
+		ARG 4 filledBucketStack
```
alcatrazEscapee, line 10: *very* inconsistient use of `fooStack`, `fooItemStack`, `fooItem` for item stack arguments. Mojang uses `stack` and `item` both, but depending on context. Pick the one that makes the most sense in this context and use it. Notably, `itemStack` is probably redundant.

data/net/minecraft/core/cauldron/CauldronInteraction.mapping
```diff
@@ -0,0 +1,128 @@
+CLASS net/minecraft/core/cauldron/CauldronInteraction net/minecraft/core/cauldron/CauldronInteraction
+	COMMENT The maps in this class are for each vanilla cauldron block. The EMPTY map handles everything, that can be done with an empty cauldron, the map itself is not empty.
+	METHOD addDefaultInteractions (Ljava/util/Map;)V
+		ARG 0 cauldronInteractionMap
+	METHOD emptyBucket (Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/sounds/SoundEvent;)Lnet/minecraft/world/InteractionResult;
+		ARG 0 level
+		ARG 1 pos
+		ARG 2 player
+		ARG 3 interactionHand
```
alcatrazEscapee, line 9: `hand`.

data/net/minecraft/core/cauldron/CauldronInteraction.mapping
```diff
@@ -0,0 +1,128 @@
...
+		ARG 4 filledBucketStack
+		ARG 5 newBlockState
```
alcatrazEscapee, line 11: `newState`

data/net/minecraft/core/cauldron/CauldronInteraction.mapping
```diff
@@ -0,0 +1,128 @@
...
+		ARG 6 soundEvent
+	METHOD fillBucket (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;Lnet/minecraft/world/item/ItemStack;Ljava/util/function/Predicate;Lnet/minecraft/sounds/SoundEvent;)Lnet/minecraft/world/InteractionResult;
+		ARG 0 blockState
```
alcatrazEscapee, line 14: `state`

data/net/minecraft/core/cauldron/CauldronInteraction.mapping
```diff
@@ -0,0 +1,128 @@
+CLASS net/minecraft/core/cauldron/CauldronInteraction net/minecraft/core/cauldron/CauldronInteraction
+	COMMENT The maps in this class are for each vanilla cauldron block. The EMPTY map handles everything, that can be done with an empty cauldron, the map itself is not empty.
```
alcatrazEscapee, line 2: Could also say "The EMPTY map is also not a kitchen sink" because that's true, and equally as redundant. What do these maps *do*?

> This class is responsible for item interactions with the vanilla cauldron block. Each `{@code Map}` will be queried for interactions when the cauldron is filled with the respective fluid.

data/net/minecraft/core/cauldron/CauldronInteraction.mapping
```diff
@@ -0,0 +1,128 @@
...
+		ARG 6 soundEvent
+	METHOD fillBucket (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;Lnet/minecraft/world/item/ItemStack;Ljava/util/function/Predicate;Lnet/minecraft/sounds/SoundEvent;)Lnet/minecraft/world/InteractionResult;
+		ARG 0 blockState
+		ARG 1 level
+		ARG 2 pos
+		ARG 3 player
+		ARG 4 hand
+		ARG 5 emptyBucketItem
+		ARG 6 filledBucketItem
+		ARG 7 filledCauldronPredicate
```
alcatrazEscapee, line 21: This is not at all what this actually is - it's a test for if the bucket is able to be filled, passing in the cauldron state. Nor is calling it `Predicate` particularly useful. I'd go with `canFillBucket` but not sold on it.

data/net/minecraft/core/cauldron/CauldronInteraction.mapping
```diff
@@ -0,0 +1,128 @@
...
+		ARG 8 soundEvent
+	METHOD interact (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;)Lnet/minecraft/world/InteractionResult;
+		ARG 1 blockState
```
alcatrazEscapee, line 24: Again, more `blockState` -> `state` and `itemStack` -> `stack`

data/net/minecraft/core/cauldron/CauldronInteraction.mapping
```diff
@@ -0,0 +1,128 @@
...
+		ARG 5 bucketItemStack
+	METHOD lambda$bootStrap$2 (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/entity/player/Player;Lnet/minecraft/world/InteractionHand;Lnet/minecraft/world/item/ItemStack;)Lnet/minecraft/world/InteractionResult;
+		ARG 0 blockState
+		ARG 1 level
+		ARG 2 pos
+		ARG 3 player
+		ARG 4 hand
+		ARG 5 potionItemStack
+	METHOD lambda$bootStrap$3 (Lnet/minecraft/world/level/block/state/BlockState;)Z
+		ARG 0 cauldronBlockState
```
alcatrazEscapee, line 45: is it important to specify in this specific case the state is of the cauldron?

data/net/minecraft/core/BlockPos.mapping
```diff
@@ -86,6 +94,7 @@ CLASS net/minecraft/core/BlockPos
 		ARG 0 pos
 		ARG 2 direction
 	METHOD randomBetweenClosed (Ljava/util/Random;IIIIIII)Ljava/lang/Iterable;
+		COMMENT @return an Iterable of blockPositions with the given amount of elements in the area.
 		ARG 0 rand
```
alcatrazEscapee, line 98: `random`

alcatrazEscapee (changes requested):

sciwhiz12:
Hello! 👋 It has been a good amount of time since this PR was last reviewed, and none of the changes requested by the reviewers have been addressed. Because of this, we will assume (unless told otherwise) that this PR has been abandoned, and will therefore close it as such. 

If you wish to resolve the requested changes and continue this PR, please leave a comment saying so and we will reopen the PR. Thank you for your contributions!

PR #39, opened 2021-07-04
Clean Up `net.minecraft.nbt`

ChampionAsh5357:
Project Phoenix Week 1:
- Clean up and map `net.minecraft.nbt`

Some additional parameters and javadocs have been added in place of those that were changed/removed.


data/net/minecraft/nbt/CompoundTag.mapping
```diff
@@ -1,109 +1,99 @@
...
+	METHOD <init> (Ljava/util/Map;)V
+		ARG 1 tags
 	METHOD contains (Ljava/lang/String;)Z
-		COMMENT Returns whether the given string has been previously stored as a key in the map.
 		ARG 1 key
 	METHOD contains (Ljava/lang/String;I)Z
-		COMMENT Returns whether the given string has been previously stored as a key in this tag compound as a particular type, denoted by a parameter in the form of an ordinal. If the provided ordinal is 99, this method will match tag types representing numbers.
+		COMMENT Returns whether the tag of the specified {@code key} is a
+		COMMENT particular {@code tagType}. If the {@code tagType} is {@code 99},
+		COMMENT all numeric tags will be checked against the type of the stored tag.
```
alcatrazEscapee: Soooo... standards decision on whether to include intentional linebreaks in javadoc comments for any given arbitrary column width? If I remember correctly we had said to only add specific line breaks when one was actually desired and leave the splitting of words up to the mapping consumer (as this does not account for things like indentation of the original javadoc, etc.)

In any case, I'd think this is all one line, or two (one per sentence), no?

data/net/minecraft/nbt/CompoundTag.mapping
```diff
@@ -1,109 +1,99 @@
...
 	METHOD hasUUID (Ljava/lang/String;)Z
 		ARG 1 key
+	METHOD lambda$static$0 (Lcom/mojang/serialization/Dynamic;)Lcom/mojang/serialization/DataResult;
+		ARG 0 dynamic
+	METHOD lambda$static$1 (Lnet/minecraft/nbt/CompoundTag;)Lcom/mojang/serialization/Dynamic;
+		ARG 0 compoundTag
 	METHOD merge (Lnet/minecraft/nbt/CompoundTag;)Lnet/minecraft/nbt/CompoundTag;
-		COMMENT Deep copies all the tags of {@code other} into this tag, then returns itself.
+		COMMENT Copies all the tags of {@code other} into this tag, then returns itself.
+		COMMENT @see #copy()
```
alcatrazEscapee, line 58: Another standards point worth clarifying: this is a reference to it's own class, so it does not need to be fully qualified. Is this correct?

- IF the goal is to provide cross-version migration ability of javadoc links, then this is not useful
- IF the goal is to make javadoc links work regardless of the presence of a particular import statement, then this is fine.

@ standards people (oh look it's me) thoughts?

alcatrazEscapee (commented):
okay I was going to look through this with intent to review but hit two standards points I think might be in the best interest to clear up / clarify.

cc rest of standards team aka @ChampionAsh5357 @SizableShrimp

data/net/minecraft/nbt/CompoundTag.mapping
ChampionAsh5357: I forgot about this. Technically it should all be on one line or on multiple lines. I was just formatting so I could read while I was mapping. I'll fix this.

data/net/minecraft/nbt/CompoundTag.mapping
ChampionAsh5357: That would be my understanding. The reason I use this method is that how copying works is applied based on the description within the javadoced method. The goal here I would say is to make the link work without an import statement.

Lanse505 (approved):

alcatrazEscapee (approved):

PR #42, opened 2021-07-08
Project Phoenix: net minecraft core

diesieben07:


data/net/minecraft/core/Vec3i.mapping
```diff
@@ -51,23 +51,20 @@ CLASS net/minecraft/core/Vec3i
...
 		ARG 1 axis
 	METHOD getX ()I
-		COMMENT Gets the X coordinate.
 	METHOD getY ()I
-		COMMENT Gets the Y coordinate.
 	METHOD getZ ()I
-		COMMENT Gets the Z coordinate.
 	METHOD multiply (I)Lnet/minecraft/core/Vec3i;
-		ARG 1 scalar
+		ARG 1 factor
```
jaredlll08:
```suggestion
		ARG 1 scalar
```

jaredlll08 (changes requested):
As discussed this should be scalar.

The rest looks perfect to me

data/net/minecraft/core/Direction.mapping
```diff
@@ -6,17 +6,27 @@ CLASS net/minecraft/core/Direction
 	FIELD data3d I
 		COMMENT Ordering index for D-U-N-S-W-E
 	FIELD normal Lnet/minecraft/core/Vec3i;
-		COMMENT Normalized Vector that points in the direction of this Facing
+		COMMENT Normalized Vector that points in the direction of this Direction
```
alcatrazEscapee:
```suggestion
		COMMENT Normalized vector that points in the direction of this Direction
```
Unless `Vector` is referencing a specific class name.

data/net/minecraft/core/Direction.mapping
```diff
@@ -6,17 +6,27 @@ CLASS net/minecraft/core/Direction
...
+		COMMENT @return the Direction specified by the given name or null if no such Direction exists
 		ARG 0 name
 	METHOD from2DDataValue (I)Lnet/minecraft/core/Direction;
-		COMMENT Gets the EnumFacing corresponding to the given horizontal index (0-3). Out of bounds values are wrapped around. The order is S-W-N-E.
+		COMMENT @return the Direction corresponding to the given horizontal index (0-3). Out of bounds values are wrapped around. The order is S-W-N-E.
+		COMMENT @see #get2DDataValue
 		ARG 0 horizontalIndex
 	METHOD from3DDataValue (I)Lnet/minecraft/core/Direction;
-		COMMENT Gets the EnumFacing corresponding to the given index (0-5). Out of bounds values are wrapped around. The order is D-U-N-S-W-E.
+		COMMENT @return the {@code Direction} corresponding to the given index (0-5). Out of bounds values are wrapped around. The order is D-U-N-S-W-E.
```
alcatrazEscapee, line 28: Less of a request more of a comment. PR has inconsistent use of `{@code Foo}` vs. `Foo` when referring to specific classes. e.g. `Direction` vs. `{@code Direction}` vs. `direction` (lowercase being a reference to the word rather than the class).

alcatrazEscapee (changes requested):
One minor grammar tweak and a nitpick / question about javadoc style. All the rest looks good to me.

data/net/minecraft/core/Direction.mapping
diesieben07: That is true and I am not sure which we should go for. I don't think the answer for which one we want is always the same either.

alcatrazEscapee: In all honesty, me neither. I was thinking of this mostly from a standards PoV, but not sure if we *want* to enforce this, one way or the other. The only thing I'd say would be good is to make the distinction between i.e. `BlockPos` and `block position` - one refers to the class, the other to the concept.

alcatrazEscapee (dismissed):

jaredlll08 (dismissed):

jaredlll08 (approved):

alcatrazEscapee (approved):

PR #43, opened 2021-07-08
Fix parameter name offsets in ClientboundCommandsPacket.mapping

diesieben07:
This is another case where there was an extra parameter in a method, which threw things off. It slipped through the radar with my previous network PR, sorry.


sciwhiz12 (approved):

sciwhiz12:
As the errors fixed by this PR are causing the validation check in other PRs to fail, I am merging this in on the basis of small-enough fixes for validation errors which get introduced into the main branches should be fixed as soon as possible.

PR #44, opened 2021-07-08
Project phoenix: n/m/world/level/levelgen/carver

alcatrazEscapee:
Also because this is my happy place I went and properly mapped all the carver parameters, and added some basic conceptual level javadocs.

Minor quibbles:
 - `biomeAccessor`: Since carvers don't generally need to query the biome (which is a non-trivial operation in terms of time spent), and they are invoked *a lot*, they pass in a `Function<BlockPos, Biome>` everywhere. Which essentially just is `world::getBiome`. I've went ahead and called these all `biomeAccessor` but if anyone has better ideas for that particular name I'm all ears.

### DIFF
--- a/data/net/minecraft/world/level/levelgen/carver/WorldCarver.mapping
+++ b/data/net/minecraft/world/level/levelgen/carver/WorldCarver.mapping
@@ -1,34 +1,64 @@
 CLASS net/minecraft/world/level/levelgen/carver/WorldCarver
 	METHOD <init> (Lcom/mojang/serialization/Codec;)V
 		ARG 1 codec
+	METHOD canReach (Lnet/minecraft/world/level/ChunkPos;DDIIF)Z
+		ARG 0 chunkPos
+		ARG 1 x
+		ARG 3 z
+		ARG 5 branchIndex
+		ARG 6 branchCount
+		ARG 7 width
 	METHOD canReplaceBlock (Lnet/minecraft/world/level/block/state/BlockState;)Z
 		ARG 1 state
 	METHOD canReplaceBlock (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/block/state/BlockState;)Z
 		ARG 1 state
 		ARG 2 aboveState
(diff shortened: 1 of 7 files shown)


Lanse505 (dismissed):

data/net/minecraft/world/level/levelgen/carver/WorldCarver.mapping
```diff
@@ -1,34 +1,64 @@
...
+		ARG 6 branchCount
+		ARG 7 width
 	METHOD canReplaceBlock (Lnet/minecraft/world/level/block/state/BlockState;)Z
 		ARG 1 state
 	METHOD canReplaceBlock (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/block/state/BlockState;)Z
 		ARG 1 state
 		ARG 2 aboveState
 	METHOD carve (Lnet/minecraft/world/level/levelgen/carver/CarvingContext;Lnet/minecraft/world/level/levelgen/carver/CarverConfiguration;Lnet/minecraft/world/level/chunk/ChunkAccess;Ljava/util/function/Function;Ljava/util/Random;Lnet/minecraft/world/level/levelgen/Aquifer;Lnet/minecraft/world/level/ChunkPos;Ljava/util/BitSet;)Z
+		COMMENT Carves the given chunk with caves that originate from the given {@code chunkPos}.
+		COMMENT This method is invoked 289 times in order to generate each chunk (once for every position in an 8x8 chunk radius centered around the target chunk).
```
jaskarth: I think you mean 17x17? since 17^2 is 289 and the code iterates from -8 to 8 on both axes.

data/net/minecraft/world/level/levelgen/carver/WorldCarver.mapping
```diff
@@ -1,34 +1,64 @@
...
+		COMMENT Carves a single block, replacing it with the appropiate state if possible, and handles replacing exposed dirt with grass.
 		ARG 1 context
 		ARG 2 config
 		ARG 3 chunk
+		ARG 4 biomeAccessor
 		ARG 5 carvingMask
 		ARG 6 random
+		ARG 7 pos
+			COMMENT The position to carve at. The method does not mutate this position.
+		ARG 8 mutablePos
```
jaskarth: I would name this "checkPos" since it's used to check for grass and dirt. It's not the best name but it's better than nothing haha

data/net/minecraft/world/level/levelgen/carver/WorldCarver.mapping
alcatrazEscapee: Yeah, I think I was trying to write both "17x17 area" and "8 chunk radius" and somehow merged the two xD

jaskarth: ahh lol

jaskarth (approved):

Lanse505 (approved):

PR #45, opened 2021-07-10
Update Compass and dependencies, sanitize and reformat data

sciwhiz12:
This PR does a number of toolchain-related changes. 
The following dependencies were updated: Compass to 0.3.0, Vignette to 0.2.0.16, Enigma to 1.4.0. The data is passed through the sanitization task, which also reformats the data using the improved Enigma format writer in Compass 0.3.0. The CI `afterEvaluate` block is moved and adjusted to only log a message when the flag is set.

---

### Data Sanitization

The sanitization task removed any entries for synthetic methods and fields. Viewers may notice the sheer quantity of removed methods which may be alarming, but this is normal. A forgotten detail about the metadata for inner classes caused validation and sanitation for inner classes to not be taken (see commit https://github.com/ParchmentMC/Compass/commit/ce8dd555baa1f50db8a49cd8cbb64ad67566fa15). Because the bug was fixed, validation and sanitization for inner classes properly work now, which caused the large amount of removed entries.

> The reason why sanitization needs to happen is because of a quirk of the SRG intermediate used when migrating the data from 1.16.5 to 1.17. SRG assigns both the actual method and its synthetic bouncer the same SRG ID and name, which causes Compass to write the same data for both. As a workaround, the sanitization task was introduced, to be ran during migration to clear out the entries for these synthetic methods and fields.

### Improved Enigma Writing

The improved Enigma format writer in Compass v0.3.0 follows the sorting behavior of the actual Enigma library closer than the previous code. Previously, it was assumed that classes were sorted based on lexicograpical order of their names. However, closer inspection of the code reveals that classes are sorted first by the length of their class names, and then by lexicographical order for same-length class names.

> The reason why classes are sorted as such in Enigma is because of the fundamental design of Enigma. Enigma works on the assumption that there are two names for any entry: the _obfuscated_ name, and the _deobfuscated_ name. Enigma sorts class entries by the obfuscated name. The sorting is optimized for obfuscated names which have a consistent prefix for types (class, method, field, etc.) and have a numerical suffix (for example, using Intermediary would make the 'obfuscated name' of a class potentially as `class_31415`.)
> Sorting only by lexicographical order would (for example) put `class_12345` before `class_35`, which would be confusing to navigate both in the Enigma GUI and in the mapping files. By sorting based on the length of the obfuscated name first then lexicographically, then the order of the numerical suffixes is preserved (in that, when viewed via GUI or in file, it would be in sequential order according to the numerical suffix).

---

## Review Note
This PR will be merged once I have done a quick review pass over the data for anything majorly out-of-place. This PR is under the purview of toolchain-related changes, and is not subject to the normal PR review process for contributions to the mappings data.


sciwhiz12:
My one-pass review is completed, and I've found nothing majorly out-of-place. I will merge this PR within a short while.

Compass is now at v0.3.2, which includes a new `sanitizeData` task to complement the `sanitizeStagingData` task. As now noted in the contributing guide, it is recommended to run the `sanitizeData` task if the validation task reports any errors before the contributor tries manually resolving the errors.

PR #54, opened 2021-07-27
Phoenix: levelgen/features and levelgen/configurations

alcatrazEscapee:
Again, lots of mapping and cleaning up what (little) cruft there was. Yes I did map all parameters in `IcebergFeature`. Yes that did require carefully browsing the wikipedia article for "Ellipse". No I am not confident they're all right but I've seen nobody braver enough to actually map that absolute unit of a class.

oop


data/net/minecraft/world/level/levelgen/feature/DripstoneUtils.mapping
```diff
@@ -3,9 +3,16 @@ CLASS net/minecraft/world/level/levelgen/feature/DripstoneUtils
 		ARG 0 direction
 		ARG 1 height
 		ARG 2 mergeTip
+		ARG 3 blockSetter
 	METHOD createPointedDripstone (Lnet/minecraft/core/Direction;Lnet/minecraft/world/level/block/state/properties/DripstoneThickness;)Lnet/minecraft/world/level/block/state/BlockState;
 		ARG 0 direction
 		ARG 1 dripstoneThickness
+	METHOD getDripstoneHeight (DDDD)D
+		COMMENT The formula used to control dripstone columns radius. <a href="https://twitter.com/henrikkniberg/status/1334180031900360707">Henrik tweeted about it!</a>
```
sciwhiz12: Could also have been a `@see <a href="...">Tweet by @henrikkniberg, "..."</a>`.

data/net/minecraft/world/level/levelgen/feature/Feature.mapping
```diff
@@ -2,11 +2,17 @@ CLASS net/minecraft/world/level/levelgen/feature/Feature
 	METHOD <init> (Lcom/mojang/serialization/Codec;)V
 		ARG 1 codec
 	METHOD checkNeighbors (Ljava/util/function/Function;Lnet/minecraft/core/BlockPos;Ljava/util/function/Predicate;)Z
+		COMMENT @return true if any of the six adjacent blocks to the block at {@code pos} return {@code true} to the given filter.
+		ARG 0 adjacentStateAccessor
+			COMMENT An accessor for the adjacent blocks. Essentially {@code level::getBlockState}.
```
sciwhiz12, line 7: It is not necessarily an accessor for _adjacent_ blocks, just an accessor for block states given a block position.

data/net/minecraft/world/level/levelgen/feature/Feature.mapping
```diff
@@ -2,11 +2,17 @@ CLASS net/minecraft/world/level/levelgen/feature/Feature
...
 		ARG 1 pos
 		ARG 2 filter
 	METHOD configured (Lnet/minecraft/world/level/levelgen/feature/configurations/FeatureConfiguration;)Lnet/minecraft/world/level/levelgen/feature/ConfiguredFeature;
 		ARG 1 config
 	METHOD isAdjacentToAir (Ljava/util/function/Function;Lnet/minecraft/core/BlockPos;)Z
+		COMMENT @return true if any of the six adjacent blocks to the block at {@code pos} are air.
+		ARG 0 adjacentStateAccessor
+			COMMENT An accessor for the adjacent blocks. Essentially {@code level::getBlockState}.
```
sciwhiz12, line 15: Same comment as above, it is not necessarily an accessor for _adjacent_ blocks.

data/net/minecraft/world/level/levelgen/feature/Feature.mapping
```diff
@@ -20,11 +26,24 @@ CLASS net/minecraft/world/level/levelgen/feature/Feature
...
+		ARG 0 configuredFeature
 	METHOD markAboveForPostProcessing (Lnet/minecraft/world/level/WorldGenLevel;Lnet/minecraft/core/BlockPos;)V
+		COMMENT If the above two blocks are not air, marks them for post processing.
+		COMMENT This is used to prevent floating grass during the generation of features that carve blocks out of the terrain, after other plant-like blocks have generated (such as lake features).
 		ARG 1 level
 		ARG 2 basePos
 	METHOD place (Lnet/minecraft/world/level/levelgen/feature/FeaturePlaceContext;)Z
-		ARG 1 featurePlaceContext
+		COMMENT Places the given feature at the given location.
+		COMMENT During world generation, features are provided with a 3x3 region of chunks that they can safely generate into.
```
sciwhiz12: The 3x3 region of chunks referenced here is centered around the block position within the place context provided, right? It may be a good idea to note that here.

data/net/minecraft/world/level/levelgen/feature/OreFeature.mapping
```diff
@@ -1,6 +1,7 @@
 CLASS net/minecraft/world/level/levelgen/feature/OreFeature
 	METHOD canPlaceOre (Lnet/minecraft/world/level/block/state/BlockState;Ljava/util/function/Function;Ljava/util/Random;Lnet/minecraft/world/level/levelgen/feature/configurations/OreConfiguration;Lnet/minecraft/world/level/levelgen/feature/configurations/OreConfiguration$TargetBlockState;Lnet/minecraft/core/BlockPos$MutableBlockPos;)Z
 		ARG 0 state
+		ARG 1 adjacentStateAccessor
```
sciwhiz12, line 4: As noted for `Feature#isAdjacentToAir`, this is not necessarily an accessor for _adjacent_ states but is a general accessor for position to state (which is used for checking adjacent positions).

data/net/minecraft/world/level/levelgen/structure/BoundingBox.mapping
```diff
@@ -9,8 +9,10 @@ CLASS net/minecraft/world/level/levelgen/structure/BoundingBox
 	METHOD <init> (Lnet/minecraft/core/BlockPos;)V
 		ARG 1 pos
 	METHOD encapsulate (Lnet/minecraft/core/BlockPos;)Lnet/minecraft/world/level/levelgen/structure/BoundingBox;
+		COMMENT Expands this box to be at least large enough to contain {@coe pos}
```
sciwhiz12: Typo:
```suggestion
		COMMENT Expands this box to be at least large enough to contain {@code pos}
```

data/net/minecraft/world/level/levelgen/structure/BoundingBox.mapping
```diff
@@ -21,35 +23,44 @@ CLASS net/minecraft/world/level/levelgen/structure/BoundingBox
...
-		ARG 1 structurebb
+		COMMENT @return true if {@code box} intersects this box.
+		ARG 1 box
 	METHOD isInside (Lnet/minecraft/core/Vec3i;)Z
-		COMMENT Checks if given Vec3i is inside of StructureBoundingBox
-		ARG 1 vec
+		COMMENT @return true if the bounding box contains the {@code vector}
+		ARG 1 vector
 	METHOD move (III)Lnet/minecraft/world/level/levelgen/structure/BoundingBox;
+		COMMENT Translates this box by the given coordinates.
```
sciwhiz12: Possibly consider explicitly declaring the return value of this value being the same instance (to note that this modifies the values in the instance), as opposed to `#moved` which returns a new instance without modifying the original.

sciwhiz12 (changes requested):
I have a few comments (along with a typo, which is why this is a change requesting review instead of an approval/commente), especially regarding marking state accessors as explicitly _adjacent_ state accessors.

data/net/minecraft/world/level/levelgen/feature/Feature.mapping
alcatrazEscapee: > is centered around the block position within the place context provided

No.
Technicality, but it is centered on the `center` chunk position passed in the `WorldGenRegion` (which implements `WorldGenLevel`) found on the `FeaturePlaceContext`. A decorator is absolutely free to move said position outside the generating chunk. (Although in practice this rarely, if ever, happens)

alcatrazEscapee:
Re: adjacent state accessors:

`level::getBlockState` is not just an adjacent state accessor, no. Nor is the general `Function<BlockPos, BlockState>`. But, in all the cases where I named it that, it was (and all) it's used for.

For a counterexample, tree generation makes heavy use of a `BiFunction<BlockPos, BlockState>` which is very generally *not* restricted. Hence the general name `blockSetter`. Wheras in these cases, it was explicitly known that the *only* states this would be used to access are adjacent ones.

In other words, you *do not need* to pass in an accessor that accesses general states given a block position. The following code works fine as well:
```java
final Map<BlockPos, BlockState> cachedAdjacentStates = Arrays.stream(Direction.values())
    .map(d -> pos.offset(d))
    .collect(Collectors.toMap(d -> d, d -> level.getBlockState(d)));
final Function<BlockPos, BlockState> adjacentStateAccessor = cachedAdjacentStates::get;
```

sciwhiz12:
Fair enough, then.

data/net/minecraft/world/level/levelgen/structure/BoundingBox.mapping
```diff
@@ -9,8 +12,10 @@ CLASS net/minecraft/world/level/levelgen/structure/BoundingBox
 	METHOD <init> (Lnet/minecraft/core/BlockPos;)V
 		ARG 1 pos
 	METHOD encapsulate (Lnet/minecraft/core/BlockPos;)Lnet/minecraft/world/level/levelgen/structure/BoundingBox;
+		COMMENT Expands this box to be at least large enough to contain {@coxe pos}
```
sciwhiz12: Typo:
```suggestion
		COMMENT Expands this box to be at least large enough to contain {@code pos}
```

data/net/minecraft/world/level/levelgen/structure/BoundingBox.mapping
```diff
@@ -21,35 +26,44 @@ CLASS net/minecraft/world/level/levelgen/structure/BoundingBox
...
+		COMMENT @return The length of this bounding box along the y axis.
 	METHOD getZSpan ()I
-		COMMENT Get dimension of the bounding box in the z direction.
+		COMMENT @return The length of this bounding box along the z axis.
 	METHOD inflate (I)Lnet/minecraft/world/level/levelgen/structure/BoundingBox;
+		COMMENT Expands this box by a fixed {@code value} in all directions.
 		ARG 1 value
 	METHOD intersects (IIII)Z
-		COMMENT Discover if a coordinate is inside the bounding box area.
+		COMMENT @return true if this bounding box inersects the horizontal x/z region described by the min and max parameters.
```
sciwhiz12: Typo:
```suggestion
		COMMENT @return true if this bounding box intersects the horizontal x/z region described by the min and max parameters.
```

sciwhiz12 (changes requested):

sciwhiz12 (dismissed):

data/net/minecraft/world/level/levelgen/feature/BasaltColumnsFeature.mapping
```diff
@@ -1,21 +1,24 @@
...
+		ARG 1 seaLevel
 		ARG 2 pos
 	METHOD findAir (Lnet/minecraft/world/level/LevelAccessor;Lnet/minecraft/core/BlockPos$MutableBlockPos;I)Lnet/minecraft/core/BlockPos;
 		ARG 0 level
 		ARG 1 pos
-		ARG 2 height
+		ARG 2 distance
 	METHOD findSurface (Lnet/minecraft/world/level/LevelAccessor;ILnet/minecraft/core/BlockPos$MutableBlockPos;I)Lnet/minecraft/core/BlockPos;
 		ARG 0 level
 		ARG 1 height
```
Harleyoc1: This may be unnecessary but the sea level is the only type of height passed into this method. 
```suggestion
		ARG 1 seaLevel
```

data/net/minecraft/world/level/levelgen/feature/Feature.mapping
```diff
@@ -2,11 +2,17 @@ CLASS net/minecraft/world/level/levelgen/feature/Feature
 	METHOD <init> (Lcom/mojang/serialization/Codec;)V
 		ARG 1 codec
 	METHOD checkNeighbors (Ljava/util/function/Function;Lnet/minecraft/core/BlockPos;Ljava/util/function/Predicate;)Z
+		COMMENT @return true if any of the six adjacent blocks to the block at {@code pos} return {@code true} to the given filter.
```
Harleyoc1:
```suggestion
		COMMENT @return {@code true} if any of the six adjacent blocks to the block at {@code pos} return {@code true} to the given filter.
```

data/net/minecraft/world/level/levelgen/feature/IcebergFeature.mapping
```diff
@@ -1,20 +1,76 @@
...
+			COMMENT The maximum height of the iceberg.
+		ARG 3 maxRadius
+			COMMENT The maximum radius of the iceberg, at the horizontal.
 	METHOD heightDependentRadiusRound (Ljava/util/Random;III)I
 		ARG 1 random
+		ARG 2 y
+		ARG 3 height
+		ARG 4 majorAxis
+	METHOD heightDependentRadiusSteep (Ljava/util/Random;III)I
+		COMMENT Given a horizontal projection of an iceberg, defines the three dimensional extrusion by defining a radius at any given y value
```
Harleyoc1:
```suggestion
		COMMENT Given a horizontal projection of an iceberg, defines the three dimensional extrusion by defining a radius at any given y value.
```

data/net/minecraft/world/level/levelgen/feature/IcebergFeature.mapping
```diff
@@ -1,20 +1,76 @@
...
+		COMMENT The radius curve is a linear function, with a slope that is both dependent on the {@code majorAxis} and randomly varies, which results in steep conical icebergs.
+		ARG 1 random
+			COMMENT A random to use to vary the slope of the falloff curve.
+		ARG 2 y
+			COMMENT The y value to calculate a radius at
```
Harleyoc1:
```suggestion
			COMMENT The y value to calculate a radius at.
```

data/net/minecraft/world/level/levelgen/structure/BoundingBox.mapping
```diff
@@ -21,35 +26,44 @@ CLASS net/minecraft/world/level/levelgen/structure/BoundingBox
...
+		COMMENT Translates this box by the given coordinates, modifying the current box.
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
 	METHOD move (Lnet/minecraft/core/Vec3i;)Lnet/minecraft/world/level/levelgen/structure/BoundingBox;
-		ARG 1 vec
+		COMMENT Translates this box by the given vector, modifying the current box.
+		ARG 1 vector
 	METHOD moved (III)Lnet/minecraft/world/level/levelgen/structure/BoundingBox;
+		COMMENT @return A new bounding box equal to this box, translated by the given coordinates
```
Harleyoc1: A few more minor things in this file like missing `@code` tags and full stops (periods).

Harleyoc1 (commented):
A few minor suggestions. I'm unsure on how strict the reviews should be on punctuation and grammar but I've put them in just in case (would appreciate some clarification on this).

alcatrazEscapee:
Re: explicit grammar (i.e. should we allow `@param foo  the foo` or `@param foo The foo.` or both? Well... we don't have an explicit standard set for that yet. It is something that we may come to an agreement / contract on in the future.

Harleyoc1 (approved):

Harleyoc1:
All right. I'll ignore stuff like that for now then, though it would be better an agreement is come to sooner than later to minimise any corrections if they end up being required.

sciwhiz12 (approved):
On the note of "`@param foo The foo.` vs `@param foo the foo`", I prefer following [Oracle's style](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#@param): `@param foo the foo` for this specific situation. We still need to agree to a style guide set though, for consistency.

PR #48, opened 2021-07-23
[Project: Phoenix]: Clean-up and Mapping of net/minecraft/world/item and it's sub-packages

Lanse505:
Cleaning-Up and Mapping the following packages:
```
net/minecraft/world/food
net/minecraft/world/item
net/minecraft/world/item/context
net/minecraft/world/item/crafting
net/minecraft/world/item/enchantment
net/minecraft/world/level/saveddata
net/minecraft/world/phys
net/minecraft/world/phys/shapes
```

PR #55, opened 2021-07-30
Parameter mappings for `net.minecraft.commands`. 

Harleyoc1:
Added and updated parameters in `net.minecraft.commands` and all sub-packages. Also updated one or two Javadocs but nothing significant. 

I should add that due to Enigma's mishandling of generics, synthetic method parameters were originally mapped. I've changed all to map the source method instead, although Enigma doesn't seem to recognise them? I have verified that the descriptors match those from the bytecode, they're just not showing up in Enigma.


data/net/minecraft/commands/BrigadierExceptions.mapping
```diff
@@ -0,0 +1,43 @@
...
+		ARG 1 value
+	METHOD lambda$static$5 (Ljava/lang/Object;Ljava/lang/Object;)Lcom/mojang/brigadier/Message;
+		ARG 0 max
+		ARG 1 value
+	METHOD lambda$static$6 (Ljava/lang/Object;Ljava/lang/Object;)Lcom/mojang/brigadier/Message;
+		ARG 0 min
+		ARG 1 value
+	METHOD lambda$static$7 (Ljava/lang/Object;Ljava/lang/Object;)Lcom/mojang/brigadier/Message;
+		ARG 0 max
+		ARG 1 value
```
sciwhiz12: For these lambdas that have two arguments of `min`/`max` and `value`, which belong to the `argument.<number>.low/big` set of translation keys, the arguments seem to be flipped.

As in, the first lambda argument for the minimum/maximum is passed as the second parameter for translation which, according to the `en_us` language file, is for the actual number value parsed. Same with the second lambda argument (the actual value), passed as the first parameter for translation which is for the bound.

For example:
```java
(min, value) -> new TranslatableComponent("argument.double.low", value, min)
```
```json
"argument.double.low": "Double must not be less than %s, found %s"
```

data/net/minecraft/commands/CommandFunction.mapping
```diff
@@ -0,0 +1,37 @@
+CLASS net/minecraft/commands/CommandFunction
+	METHOD <init> (Lnet/minecraft/resources/ResourceLocation;[Lnet/minecraft/commands/CommandFunction$Entry;)V
+		ARG 1 id
+		ARG 2 entries
+	METHOD fromLines (Lnet/minecraft/resources/ResourceLocation;Lcom/mojang/brigadier/CommandDispatcher;Lnet/minecraft/commands/CommandSourceStack;Ljava/util/List;)Lnet/minecraft/commands/CommandFunction;
+		ARG 0 resourceLocation
```
sciwhiz12: Should be `id`, as the parameter is passed as the first argument to the `CommandFunction` constructor.

data/net/minecraft/commands/arguments/item/FunctionArgument.mapping
```diff
@@ -1,4 +1,24 @@
 CLASS net/minecraft/commands/arguments/item/FunctionArgument
+	METHOD getFunction (Lcom/mojang/brigadier/context/CommandContext;Lnet/minecraft/resources/ResourceLocation;)Lnet/minecraft/commands/CommandFunction;
+		ARG 0 context
+		ARG 1 resourceLocation
+	METHOD getFunctionOrTag (Lcom/mojang/brigadier/context/CommandContext;Ljava/lang/String;)Lcom/mojang/datafixers/util/Pair;
+		ARG 0 context
+		ARG 1 name
+	METHOD getFunctionTag (Lcom/mojang/brigadier/context/CommandContext;Lnet/minecraft/resources/ResourceLocation;)Lnet/minecraft/tags/Tag;
+		ARG 0 context
+		ARG 1 resourceLocation
```
sciwhiz12: A better name for the `ResourceLocation` parameters would be `id` or `functionId`, instead of the generic `resourceLocation`.

sciwhiz12 (changes requested):

data/net/minecraft/commands/BrigadierExceptions.mapping
Harleyoc1: Ah yes, I was paying attention to the order of the lambda parameters not noticing that they were being passed in the other way around.

sciwhiz12: Gotta love those subtle moments of mojank biting back. :D

sciwhiz12 (dismissed):

alcatrazEscapee:
> I have verified that the descriptors match those from the bytecode, they're just not showing up in Enigma.

From what we know Enigma is mapping the synthetic bouncer method, not the actual method. Our synthetic method validator rejects these (although you can actually hand-fix it, I've done that once and it worked). Proper solution TBD.

Harleyoc1:
Oh yeah I know that, I manually fixed it here too; I was saying Enigma still doesn't recognise these despite them matching the descriptor of the actual method.

data/net/minecraft/world/phys/shapes/BooleanOp.mapping
```diff
@@ -0,0 +1,52 @@
+CLASS net/minecraft/world/phys/shapes/BooleanOp
```
alcatrazEscapee: This class is already mapped by https://github.com/ParchmentMC/Parchment/pull/48 and I also left a comment https://github.com/ParchmentMC/Parchment/pull/48/files#r676217878 (I think left + right are better than first + second given the connotations of left hand side <op> right hand side

To avoid unnecessary conflicts I think the easiest would be to drop this from this PR.

alcatrazEscapee (commented):
Seeing as https://github.com/ParchmentMC/Parchment/pull/48 already has mappings for the `phys/shapes` I'd preferentially drop them from this PR to avoid creating unnecessary merge conflicts. But otherwise nothing wrong with the commands section.

data/net/minecraft/world/phys/shapes/BooleanOp.mapping
Harleyoc1: I didn't intend to commit these here, I'll just drop all `net.minecraft.world.phys` mappings from this PR.

alcatrazEscapee (approved):

sciwhiz12 (approved):

PR #57, opened 2021-08-03
Parameter mappings and Javadoc for `net.minecraft.resources`.

Harleyoc1:
Maps all parameters in `net.minecraft.resources`. I also wrote some Javadoc for the `ResourceLocation` and `ResourceKey` classes and some of their methods.


Lanse505 (dismissed):
Looks good to me

data/net/minecraft/resources/ResourceKey.mapping
```diff
@@ -1,14 +1,45 @@
 CLASS net/minecraft/resources/ResourceKey
+	COMMENT A class representing an immutable key for a resource, in terms of the name of its
```
sciwhiz12: Small bit of pedantry (ignore at your discretion): The first part of the sentence can be shortened to `An immutable key for a resource,`, since the name of the class and its definition makes it clear that it is a class.

data/net/minecraft/resources/ResourceKey.mapping
```diff
@@ -1,14 +1,45 @@
...
+	COMMENT
+	COMMENT <p>{@link net.minecraft.core.Registry} uses this to return resource keys for registry
+	COMMENT objects via {@link net.minecraft.core.Registry#getResourceKey(Object)}. It also uses
+	COMMENT this class to store its name, with the parent registry name set to {@code minecraft:root}.
+	COMMENT When used in this way it is usually referred to as a "registry key".</p>
+	COMMENT
+	COMMENT @param <T> The type of the resource represented by this {@code ResourceKey}, or the type of the registry if it is a registry key.
+	COMMENT @see net.minecraft.resources.ResourceLocation
+	FIELD location Lnet/minecraft/resources/ResourceLocation;
+		COMMENT Holds the location of the resource within the registry.
```
sciwhiz12: Small bit of pedantry (ignore at your discretion): The "Holds" word is unnecessary, as a field has no other purpose than to hold values. So, this could have been shortened to `The location of the resource within the registry`.

data/net/minecraft/resources/ResourceLocation.mapping
```diff
@@ -1,29 +1,91 @@
 CLASS net/minecraft/resources/ResourceLocation
+	COMMENT A class representing the immutable location of a resource, in terms of a path and namespace.
```
sciwhiz12: Same bit of pedantry here, first part could be shortened down to `An immutable location of a resource,`.

sciwhiz12 (dismissed):
Looks good to me. I made some small pedantic comments (which I will immediately mark as resolved), just to note possible ways to shorten the javadocs without sacrificing readability. I don't need those changes to be done, but it would be good to keep them in mind in the future.

sciwhiz12 (approved):

alcatrazEscapee (approved):

PR #62, opened 2021-08-09
Phoenix levelgen, chunk, and biome.

alcatrazEscapee:
The rest of the `levelgen` package, reviewed, parameter-mapped, and documented in key places (aka where I knew enough to comment and felt like it :P )

A note for reviewers: most of the red/removals were items dropped by Enigma, after noticing they were exact copies of subclass mappings. As of recent changes in Compass, that is no longer nessecary.

Also, for @SizableShrimp this PR introduces what (I believe) to be the first instance of a different parameter mapping between subclass and superclass, for `Cavifier` and `NoiseModifier`:
- Enigma doesn't recognize this at all (it will only allow you to map the superclass), and I instead manually wrote out the parameter here.

Covered:
- /world/level/biome
- /world/level/border
- /world/level/chunk
- /world/level/levelgen


data/net/minecraft/world/level/levelgen/Beardifier.mapping
```diff
@@ -1,8 +1,26 @@
 CLASS net/minecraft/world/level/levelgen/Beardifier
+	COMMENT Modifies terrain noise to be flatter near structures such as villages.
+	COMMENT It's called a beardifier since it looks like Jeb's beard, the bottom part is solid and the top part is shaved.
```
Lanse505: Okey out of pure curiosity.... Source?

Lanse505 (dismissed):

data/net/minecraft/world/level/levelgen/Beardifier.mapping
alcatrazEscapee: Uhh... source said it was a joke and I'm just dumb :P

Lanse505 (dismissed):

data/net/minecraft/world/level/chunk/ChunkSource.mapping
```diff
@@ -1,7 +1,10 @@
 CLASS net/minecraft/world/level/chunk/ChunkSource
 	METHOD gatherStats ()Ljava/lang/String;
-		COMMENT Converts the instance data to a readable string.
+		COMMENT @return A human readable string representing data about this chunk source.
 	METHOD getChunk (IILnet/minecraft/world/level/chunk/ChunkStatus;Z)Lnet/minecraft/world/level/chunk/ChunkAccess;
+		COMMENT Gets the chunk at the provided position, if it exists.
+		COMMENT If {@code load} is {@code true}, then this will force a chunk load.
```
Harleyoc1: Could be parameter doc for `load` instead.

data/net/minecraft/world/level/chunk/ChunkStatus.mapping
```diff
@@ -1,4 +1,7 @@
 CLASS net/minecraft/world/level/chunk/ChunkStatus
+	COMMENT The statuses that chunks go through during different phases of generation and loading.
+	COMMENT Each status has an asynchronous task that is completed to generate a chunk, and one to load a chunk up to that status.
+	COMMENT Chunks are generated in sequential stages, some of which rely on nearby chunks from the previous stage. To this respect, tasks define a "range" that they require chunks to be generated up to the previous stage. This is responsible for the cocentric squares seen in the chunk loading screen.
```
Harleyoc1:
```suggestion
	COMMENT Chunks are generated in sequential stages, some of which rely on nearby chunks from the previous stage. To this respect, tasks define a "range" that they require chunks to be generated up to the previous stage. This is responsible for the concentric squares seen in the chunk loading screen.
```

data/net/minecraft/world/level/chunk/ChunkStatus.mapping
```diff
@@ -74,10 +202,11 @@ CLASS net/minecraft/world/level/chunk/ChunkStatus
 			ARG 4 generator
 			ARG 5 structureManager
 			ARG 6 lightEngine
-			ARG 7 loadingFunction
-			ARG 8 chunks
+			ARG 7 task
+			ARG 8 neighboringChunks
 			ARG 9 loadingChunk
 	CLASS SimpleGenerationTask
+		COMMENT An {@link GenerationTask} which completes all work synchronously.
```
Harleyoc1:
```suggestion
		COMMENT A {@link GenerationTask} which completes all work synchronously.
```

data/net/minecraft/world/level/levelgen/Aquifer.mapping
```diff
@@ -1,19 +1,36 @@
 CLASS net/minecraft/world/level/levelgen/Aquifer
+	COMMENT Aquifers are responsible for non-sea level fluids found in terrain generation, but also managing that different aquifers don't intersect with eachother in ways that would create undesirable fluid placement.
```
Harleyoc1:
```suggestion
	COMMENT Aquifers are responsible for non-sea level fluids found in terrain generation, but also managing that different aquifers don't intersect with each other in ways that would create undesirable fluid placement.
```

data/net/minecraft/world/level/levelgen/Aquifer.mapping
```diff
@@ -1,19 +1,36 @@
...
+	COMMENT The aquifer interface itself is a modifier on a per-block basis. It computes a block state to be placed for each position in the world.
+	COMMENT <p>
+	COMMENT Aquifers work by first partitioning a single chunk into a low resolution grid. They then generate, via various noise layers, an {@link Aquifer.AquiferStatus} at each grid point.
```
Harleyoc1:
```suggestion
	COMMENT Aquifers work by first partitioning a single chunk into a low resolution grid. They then generate, via various noise layers, an {@link NoiseBasedAquifer.AquiferStatus} at each grid point.
```

data/net/minecraft/world/level/levelgen/Aquifer.mapping
```diff
@@ -1,19 +1,36 @@
...
+	COMMENT At each point, the grid cell containing that point is calculated, and then of the eight grid corners, the three closest aquifers are found, by square eucildean distance.
```
Harleyoc1:
```suggestion
	COMMENT At each point, the grid cell containing that point is calculated, and then of the eight grid corners, the three closest aquifers are found, by square euclidean distance.
```

data/net/minecraft/world/level/levelgen/Aquifer.mapping
```diff
@@ -1,19 +1,36 @@
...
 		ARG 4 noiseGeneratorSettings
 		ARG 5 sampler
+		ARG 6 minY
+		ARG 7 height
 	METHOD createDisabled (ILnet/minecraft/world/level/block/state/BlockState;)Lnet/minecraft/world/level/levelgen/Aquifer;
+		COMMENT Creates a disabled, or no-op aquifer. This will fill any open areas below sea level with the default fluid.
 		ARG 0 seaLevel
 		ARG 1 defaultFluid
+	METHOD shouldScheduleFluidUpdate ()Z
+		COMMENT Returns true there should be a fluid update scheduled - due to a fluid block being placed in a possibly unsteady position - at the last position passed into {@link #computeState}.
```
Harleyoc1:
```suggestion
		COMMENT Returns {@code true} if there should be a fluid update scheduled - due to a fluid block being placed in a possibly unsteady position - at the last position passed into {@link #computeState}.
```

data/net/minecraft/world/level/levelgen/Cavifier.mapping
```diff
@@ -1,4 +1,48 @@
 CLASS net/minecraft/world/level/levelgen/Cavifier
 	METHOD <init> (Lnet/minecraft/world/level/levelgen/RandomSource;I)V
-		ARG 1 randomSource
+		ARG 1 random
```
Harleyoc1, line 3: I think `randomSource` was probably better (at least from what the class name suggests: that it's a source of random, not a random object itself)?

data/net/minecraft/world/level/levelgen/Column.mapping
```diff
@@ -1,4 +1,53 @@
 CLASS net/minecraft/world/level/levelgen/Column
+	COMMENT A representation of an interger valued interval. Can be either unbounded, bounded above or below, or bounded.
```
Harleyoc1: Second sentence could do with rewording. `either` is used before more than two things and it's unclear how being "bounded above or below" and simply "bounded" is different.

data/net/minecraft/world/level/levelgen/NoiseInterpolator.mapping
```diff
@@ -1,4 +1,6 @@
 CLASS net/minecraft/world/level/levelgen/NoiseInterpolator
+	COMMENT An abstraction for the "perlerp" (itself a portmanteau of "perlin", from Perlin noise, and "lerp", indicating linear interpolation.) function used in Minecraft's noise generation.
```
Harleyoc1:
```suggestion
	COMMENT An abstraction for the "perlerp" (itself a portmanteau of "perlin", from Perlin noise, and "lerp", indicating linear interpolation) function used in Minecraft's noise generation.
```

data/net/minecraft/world/level/levelgen/NoiseInterpolator.mapping
```diff
@@ -1,4 +1,6 @@
...
+	COMMENT Tri-linear interpolation avoids sampling noise at each position, and instead samples cell corners, and linearly interpolates values between said corners. This manages said interpolation, in a way friendly to iteration to avoid unnesecary computations.
```
Harleyoc1:
```suggestion
	COMMENT Tri-linear interpolation avoids sampling noise at each position, and instead samples cell corners, and linearly interpolates values between said corners. This manages said interpolation, in a way friendly to iteration to avoid unnecessary computations.
```

data/net/minecraft/world/level/levelgen/NoiseSampler.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/world/level/levelgen/NoiseSampler
+	COMMENT Samples the base terrain density noise, using the biomes depth and scale factors.
```
Harleyoc1:
```suggestion
	COMMENT Samples the base terrain density noise, using the biome's depth and scale factors.
```

data/net/minecraft/world/level/levelgen/NoiseSampler.mapping
```diff
@@ -9,5 +10,27 @@ CLASS net/minecraft/world/level/levelgen/NoiseSampler
 		ARG 7 islandNoise
 		ARG 8 depthNoise
 		ARG 9 caveNoiseModifier
+	METHOD applySlide (DI)D
+		COMMENT Applies a slide factor to the density noise, which interpolates the edges of the world towards a target value.
+		ARG 1 noise
+		ARG 3 y
+	METHOD computeInitialDensity (IDDD)D
+		COMMENT Computes an initial density for the noise, which controls the overall style of the terrain.
+		COMMENT For example, in the overworld, this density is inversely poportional to the y value, and results in terrain that is solid at the bottom and air at the top.
```
Harleyoc1:
```suggestion
		COMMENT For example, in the overworld, this density is inversely proportional to the y value, and results in terrain that is solid at the bottom and air at the top.
```

data/net/minecraft/world/level/levelgen/RandomSource.mapping
```diff
@@ -1,4 +1,10 @@
 CLASS net/minecraft/world/level/levelgen/RandomSource
+	COMMENT A basic interface for random number generation. This mirrors the same methods in {@link java.util.Random}, however it does not make any gaurentee that these are threadsafe, unlike {@code Random}.
```
Harleyoc1:
```suggestion
	COMMENT A basic interface for random number generation. This mirrors the same methods in {@link java.util.Random}, however it does not make any guarantee that these are thread-safe, unlike {@code Random}.
```

data/net/minecraft/world/level/levelgen/RandomSource.mapping
```diff
@@ -1,4 +1,10 @@
...
+	COMMENT The notable difference is that {@link #setSeed(long)} is not {@code synchronized} and should not be accessed from multiple threads.
+	COMMENT The documentation for each individual method can be assumed to be otherwise the same as the identically named method in {@link java.util.Random}.
+	COMMENT <p>
```
Harleyoc1: Unnecessary paragraph tag (rendered views normally render `@see` and the like on a new line anyway).

data/net/minecraft/world/level/levelgen/VerticalAnchor.mapping
```diff
@@ -1,4 +1,9 @@
 CLASS net/minecraft/world/level/levelgen/VerticalAnchor
+	COMMENT Represents a resolvable height value, or y coordinate, based on the world minimum and maximum height.
+	COMMENT Can take one of the following three forms:
+	COMMENT - An absolute y value ({@link Absolute}).
+	COMMENT - A height above the lowest valid y value in the level ({@link AboveBottom}).
+	COMMENT - A height below the highest valid y value in the level ({@link BelowTop}).
```
Harleyoc1: Lists without proper formatting don't show up on rendered views properly.  

```suggestion
      COMMENT <ul>
	COMMENT <li>An absolute y value ({@link Absolute}).</li>
	COMMENT <li>A height above the lowest valid y value in the level ({@link AboveBottom}).</li>
	COMMENT <li>A height below the highest valid y value in the level ({@link BelowTop}).</li>
	COMMENT </ul>
```

Harleyoc1 (changes requested):

data/net/minecraft/world/level/levelgen/Cavifier.mapping
alcatrazEscapee: Right, so I thought this was the case until I actually looked at the implementations in question.

We have:
 - `java.util.Random`: What we all agree on is a `random`.
 - `RandomSource`: Mojang's interface, which implements the exact same methods as `java.util.Random`, to the point where the following code is perfectly valid:

```java
class EzPzRandomSource extends Random implements RandomSource {}
```

- `SimpleRandomSource` is a **copy of** `java.util.Random`, which implements `RandomSource`, with one notable change: instead of being thread-safe, it throws on concurrent access. Useful for world gen.
- `WorldGenRandom` both `extends java.util.Random`, and `implements RandomSource`, meaning that in a lot of cases, a `RandomSource` *is* a `random`, despite being a subclass. This class in particular just adds some additional seeding methods used in various places.

Lanse505: Honestly in this case looking at the code, unless there is a field, that I'm missing that references it being "randomSource". 
I'd just call it "random" since the purpose it fills is to just be a "random"

data/net/minecraft/world/level/levelgen/Column.mapping
alcatrazEscapee: The difference is a mathematics convention when talking about [bounded, left bounded, right bounded, or unbounded intervals](https://en.wikipedia.org/wiki/Interval_(mathematics)#Terminology).

However yes, I shouldn't be assuming said domain-specific knowledge there.

data/net/minecraft/world/level/levelgen/Cavifier.mapping
alcatrazEscapee: Yes, that was essentially my thought process. In the absence of a specific reason for it to be a `randomSource` (i.e., a indicative field or method name, of which I have not found), it's functionally a `random`.

data/net/minecraft/world/level/chunk/ChunkStatus.mapping
alcatrazEscapee: I'll just mention this because I found it funny, I looked at this change and thought, "no, that's wrong, I'm right", then realized it's because I read that phonetically as "An at link GenerationTask" :P

alcatrazEscapee:
@Harleyoc1 fixed the typos (thanks), minus the random source change as I disagree with it (and tried to explain the reasoning in the above comment), please re-review.

Harleyoc1 (approved):
Looks good. I take your point about the random source, feel free to mark it resolved. 👍🏻

Lanse505 (approved):

PR #67, opened 2021-08-26
Fix invalid parameter mappings

SizableShrimp:
This PR fixes some invalid parameters, mainly parameters that got missed either as old data from MCP or methods made static/non-static in the update of Parchment to 1.17.

This PR had `sanitizeData` run on it using a local build of Compass with [this commit](https://github.com/SizableShrimp/Compass/commit/14f8f40576446e3ecd015dfd5b7a15698f983e6d) to catch all the invalid parameters. Some synthetic data got removed automatically as well.


data/net/minecraft/advancements/AdvancementProgress.mapping
```diff
@@ -20,12 +20,3 @@ CLASS net/minecraft/advancements/AdvancementProgress
...
 		ARG 2 requirements
-	CLASS Serializer
-		METHOD deserialize (Lcom/google/gson/JsonElement;Ljava/lang/reflect/Type;Lcom/google/gson/JsonDeserializationContext;)Ljava/lang/Object;
-			ARG 1 json
-			ARG 2 typeOfT
-			ARG 3 context
-		METHOD serialize (Ljava/lang/Object;Ljava/lang/reflect/Type;Lcom/google/gson/JsonSerializationContext;)Lcom/google/gson/JsonElement;
-			ARG 1 src
-			ARG 2 typeOfSrc
-			ARG 3 context
```
Harleyoc1: This could be made valid instead of just removing it: 

```
	CLASS Serializer
		METHOD deserialize (Lcom/google/gson/JsonElement;Ljava/lang/reflect/Type;Lcom/google/gson/JsonDeserializationContext;)Lnet/minecraft/advancements/AdvancementProgress;
			ARG 1 json
			ARG 2 typeOfT
			ARG 3 context
		METHOD serialize (Lnet/minecraft/advancements/AdvancementProgress;Ljava/lang/reflect/Type;Lcom/google/gson/JsonSerializationContext;)Lcom/google/gson/JsonElement;
			ARG 1 src
			ARG 2 typeOfSrc
			ARG 3 context
```

Harleyoc1 (changes requested):

data/net/minecraft/advancements/AdvancementProgress.mapping
Lanse505: I think those might have been removed due to enigma cleaning them up, not him removing them.
@SizableShrimp can probably confirm this

Harleyoc1: Probably, I'm just suggesting this as I had already written out the fix. The reason they were removed is that they are synthetic due to generics, so it makes sense to keep the corrected mappings.

SizableShrimp: There is no need to keep them. They should be cascaded in the exports.

Harleyoc1: Ah OK. So it automatically populates this data from dependencies on export?

SizableShrimp: Oh dependencies? Actually I doubt it. I forgot these were from gson and not a parameter mapped inside of the minecraft codebase. Then we should probably keep it but the issue is every time sanitizeData is run this will get deleted and we have to undo it each time.

Harleyoc1: sanitizeData shouldn't delete it once the actual method is mapped, it's just because the synthetic version (with `Ljava/lang/Object` rather than `Lnet/minecraft/advancements/AdvancementProgress`) was left in.

Harleyoc1: That's what the code block above is for; maybe I didn't make it clear enough. Should work once that is pasted in as I copied the descriptor from the bytecode.

SizableShrimp: Ahh okay will fix 👍

SizableShrimp: Took me a while to get back around to it, but this is now fixed.

SizableShrimp:
Should be ready for re-review now.

Harleyoc1 (dismissed):

sciwhiz12 (dismissed):

sciwhiz12 (approved):

PR #69, opened 2021-09-23
Changed all obfuscated parameters in entity class

tmvkrpxl0:
I have changed all obfuscated parameters in Entity class

### DIFF
--- a/data/net/minecraft/world/entity/Entity.mapping
+++ b/data/net/minecraft/world/entity/Entity.mapping
@@ -189,7 +189,7 @@ CLASS net/minecraft/world/entity/Entity
 	METHOD getRandomZ (D)D
 		ARG 1 scale
 	METHOD getRelativePortalPosition (Lnet/minecraft/core/Direction$Axis;Lnet/minecraft/BlockUtil$FoundRectangle;)Lnet/minecraft/world/phys/Vec3;
-		ARG 1 portalDirection
+		ARG 1 axis
 		ARG 2 rectangle
 	METHOD getRopeHoldPosition (F)Lnet/minecraft/world/phys/Vec3;
 		ARG 1 partialTicks
--- a/data/net/minecraft/world/entity/Entity.mapping
+++ b/data/net/minecraft/world/entity/Entity.mapping
@@ -1,7 +1,4 @@
 CLASS net/minecraft/world/entity/Entity
-	METHOD <init> (Lnet/minecraft/world/entity/EntityType;Lnet/minecraft/world/level/Level;)V
-		ARG 1 type
-		ARG 2 world
 	METHOD absMoveTo (DDD)V
 		ARG 1 x
 		ARG 3 y
@@ -49,9 +46,6 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 1 pose
 	METHOD canRide (Lnet/minecraft/world/entity/Entity;)Z
 		ARG 1 entity
-	METHOD causeFallDamage (FF)Z
-		ARG 1 fallDistance
-		ARG 2 multiplier
 	METHOD changeDimension (Lnet/minecraft/server/level/ServerLevel;)Lnet/minecraft/world/entity/Entity;
 		ARG 1 server
 	METHOD checkDespawn ()V
@@ -111,15 +105,8 @@ CLASS net/minecraft/world/entity/Entity
 		COMMENT Plays the {@link #getSplashSound() splash sound}, and the {@link ParticleType#WATER_BUBBLE} and {@link ParticleType#WATER_SPLASH} particles.
 	METHOD ejectPassengers ()V
 		COMMENT Dismounts all entities riding this entity from this entity.
-	METHOD equals (Ljava/lang/Object;)Z
-		ARG 1 other
 	METHOD fillCrashReportCategory (Lnet/minecraft/CrashReportCategory;)V
 		ARG 1 category
-	METHOD fillIndirectPassengers (ZLjava/util/Set;)V
-		ARG 1 checkPlayer
-		ARG 2 passengers
-	METHOD findDimensionEntryPoint (Lnet/minecraft/server/level/ServerLevel;)Lnet/minecraft/world/level/portal/PortalInfo;
-		ARG 1 world
 	METHOD getBlockExplosionResistance (Lnet/minecraft/world/level/Explosion;Lnet/minecraft/world/level/BlockGetter;Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/material/FluidState;F)F
 		COMMENT Explosion resistance of a block relative to this entity
 		ARG 1 explosion
@@ -152,10 +139,6 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 1 livingEntity
 	METHOD getEncodeId ()Ljava/lang/String;
 		COMMENT Returns the string that identifies this Entity's class
-	METHOD getExitPortal (Lnet/minecraft/server/level/ServerLevel;Lnet/minecraft/core/BlockPos;Z)Ljava/util/Optional;
-		ARG 1 world
-		ARG 2 position
-		ARG 3 isNether
 	METHOD getEyeHeight (Lnet/minecraft/world/entity/Pose;)F
 		ARG 1 pose
 	METHOD getEyeHeight (Lnet/minecraft/world/entity/Pose;Lnet/minecraft/world/entity/EntityDimensions;)F
@@ -163,10 +146,6 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 2 size
 	METHOD getEyePosition (F)Lnet/minecraft/world/phys/Vec3;
 		ARG 1 partialTicks
-	METHOD getFluidHeight (Lnet/minecraft/tags/Tag;)D
-		ARG 1 fluidTag
-	METHOD getHorizontalDistanceSqr (Lnet/minecraft/world/phys/Vec3;)D
-		ARG 0 vec3
 	METHOD getInputVector (Lnet/minecraft/world/phys/Vec3;FF)Lnet/minecraft/world/phys/Vec3;
 		ARG 0 relative
 		ARG 1 motionScaler
@@ -191,9 +170,6 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 1 scale
 	METHOD getRandomZ (D)D
 		ARG 1 scale
-	METHOD getRelativePortalPosition (Lnet/minecraft/core/Direction$Axis;Lnet/minecraft/BlockUtil$FoundRectangle;)Lnet/minecraft/world/phys/Vec3;
-		ARG 1 direction
-		ARG 2 result
 	METHOD getRopeHoldPosition (F)Lnet/minecraft/world/phys/Vec3;
 		ARG 1 partialTicks
 	METHOD getRotationVector ()Lnet/minecraft/world/phys/Vec2;
@@ -232,8 +208,6 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 1 pos
 	METHOD hasIndirectPassenger (Lnet/minecraft/world/entity/Entity;)Z
 		ARG 1 entity
-	METHOD hasPassenger (Ljava/lang/Class;)Z
-		ARG 1 entityType
 	METHOD hasPassenger (Lnet/minecraft/world/entity/Entity;)Z
 		ARG 1 entity
 	METHOD hasPermissions (I)Z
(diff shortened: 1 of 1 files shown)


data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -139,13 +152,21 @@ CLASS net/minecraft/world/entity/Entity
...
 		ARG 1 pose
 	METHOD getEyeHeight (Lnet/minecraft/world/entity/Pose;Lnet/minecraft/world/entity/EntityDimensions;)F
 		ARG 1 pose
 		ARG 2 size
 	METHOD getEyePosition (F)Lnet/minecraft/world/phys/Vec3;
 		ARG 1 partialTicks
+	METHOD getFluidHeight (Lnet/minecraft/tags/Tag;)D
+		ARG 1 fluidTag
+	METHOD getHorizontalDistanceSqr (Lnet/minecraft/world/phys/Vec3;)D
+		ARG 0 vec3
```
Harleyoc1: Abbreviations should be avoided. Generally `vector` is preferred despite the class name, or perhaps `position` (assuming the context).

Harleyoc1 (changes requested):
I suggest you review the [mappings standards](https://parchmentmc.org/docs/mappings.html) before making any more changes. 

Along with the annotated comment, most of my issues with these changes can be summed up by three points: some names are using MCP conventions where they should be in-line with official mappings (`world` should be `level`, namely), others should use surrounding field/method names, and others could and should be simplified as allowed by the standards (e.g. `blockState` -> `state`).

data/net/minecraft/world/entity/Entity.mapping
tmvkrpxl0: I see, then, what should I use for parameter names that cannot be referenced from any other parts of the code?
For example, `Object#equals(Object)` method's parameter names are always `p_equals_1_`. What name for such function would be good?
Also, Is it recommended to change parameter names of lambda functions? If so, is it allowed to use `_unused` when parameters are unused?

Harleyoc1: Either `other` or `object` are fine for the equals method, I believe. Yes, lambda functions should also be mapped. If they are never used and what they are can't be interpreted from uses then `unused` or similar should be fine.

tmvkrpxl0: Made some changes to mapping following your advice

tmvkrpxl0: Does it violate any other mapping standard now?

tmvkrpxl0:
> I suggest you review the [mappings standards](https://parchmentmc.org/docs/mappings.html) before making any more changes.
> 
> Along with the annotated comment, most of my issues with these changes can be summed up by three points: some names are using MCP conventions where they should be in-line with official mappings (`world` should be `level`, namely), others should use surrounding field/method names, and others could and should be simplified as allowed by the standards (e.g. `blockState` -> `state`).

Sorry for bothering you, but Could you re-check the change I made after last review?

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -170,6 +188,9 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 1 scale
 	METHOD getRandomZ (D)D
 		ARG 1 scale
+	METHOD getRelativePortalPosition (Lnet/minecraft/core/Direction$Axis;Lnet/minecraft/BlockUtil$FoundRectangle;)Lnet/minecraft/world/phys/Vec3;
+		ARG 1 portalDirection
```
Harleyoc1: This is an axis, not a direction. 
```suggestion
		ARG 1 axis
```

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -170,6 +188,9 @@ CLASS net/minecraft/world/entity/Entity
...
+		ARG 2 rectangle
```
Harleyoc1: This is effectively the portal's shape, though I'd avoid `portalShape` exactly since there is a class called `PortalShape` which may be misleading as this is a `FoundRectangle`. 

```suggestion
		ARG 2 portal
```

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -284,6 +310,21 @@ CLASS net/minecraft/world/entity/Entity
 	METHOD killed (Lnet/minecraft/server/level/ServerLevel;Lnet/minecraft/world/entity/LivingEntity;)V
 		ARG 1 level
 		ARG 2 killedEntity
+	METHOD lambda$collide$1 (Lnet/minecraft/world/entity/Entity;)Z
+		ARG 0 unused
```
Harleyoc1: Even though it's unused, I would be tempted to use `entity` as it would be more instantly recognisable as an entity predicate.

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -284,6 +310,21 @@ CLASS net/minecraft/world/entity/Entity
...
+		ARG 2 exitPortal
+	METHOD lambda$getDisplayName$9 (Lnet/minecraft/network/chat/Style;)Lnet/minecraft/network/chat/Style;
+		ARG 1 currentStyle
+	METHOD lambda$isInWall$2 (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/core/BlockPos;)Z
+		ARG 1 state
+		ARG 2 pos
+	METHOD lambda$move$0 (Lnet/minecraft/world/level/block/state/BlockState;)Z
+		ARG 0 loadedState
+	METHOD lambda$null$3 (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/core/BlockPos;)Z
+		ARG 2 pos
```
Harleyoc1, line 326: `lambda$null`? I suspect something has gone wrong with the mapping software here?

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -284,6 +310,21 @@ CLASS net/minecraft/world/entity/Entity
...
+	METHOD lambda$teleportTo$10 (Lnet/minecraft/server/level/ServerLevel;Lnet/minecraft/world/entity/Entity;)V
+		ARG 1 passenger
```
Harleyoc1: Given that this includes self and passengers, I would call this `currentEntity` (not `entity` to avoid conflict with local within the lambda).

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -502,6 +553,9 @@ CLASS net/minecraft/world/entity/Entity
 	METHOD setSilent (Z)V
 		COMMENT When set to true the entity will not play sounds.
 		ARG 1 isSilent
+	METHOD setSlot (ILnet/minecraft/world/item/ItemStack;)Z
+		ARG 1 slotNumber
```
Harleyoc1:
```suggestion
		ARG 1 slotIndex
```

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -502,6 +553,9 @@ CLASS net/minecraft/world/entity/Entity
...
+		ARG 2 itemStack
```
Harleyoc1:
```suggestion
		ARG 2 stack
```

Harleyoc1 (changes requested):

data/net/minecraft/world/entity/Entity.mapping
tmvkrpxl0: I guess it's lambda method in `Entity#isInWall`, one that takes BlockState and BlockPos as arguments. That is the only lambda function that takes BlockState and BlockPos in Entity class. Not sure why that happens

tmvkrpxl0: Not sure, Weird thing is, There is only one lambda method that takes BlockState and BlockPos as arguments. and that is one in `Entity#isInWall`

tmvkrpxl0: Found it, it's second lambda function in `Entity#findDimensionEntryPoint`, in `BlockUtil.getLargestRectangleAround`

data/net/minecraft/world/entity/Entity.mapping
tmvkrpxl0: Changed it to `currentEntity`

data/net/minecraft/world/entity/Entity.mapping
tmvkrpxl0: Changed it to `entity`

tmvkrpxl0:
Bump

data/net/minecraft/world/entity/Entity.mapping
Harleyoc1: This has been marked resolved but doesn't seem to have changed. Could you try to remap it or if that does the same thing manually find the identifier?

Harleyoc1: Never mind; looking at the byte code its identifier is indeed `lambda$null$4`. Seems that it calls it that because it's embedded in another lambda. My bad, I should've checked this before creating the comment.

Harleyoc1 (dismissed):

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -105,8 +108,15 @@ CLASS net/minecraft/world/entity/Entity
 		COMMENT Plays the {@link #getSplashSound() splash sound}, and the {@link ParticleType#WATER_BUBBLE} and {@link ParticleType#WATER_SPLASH} particles.
 	METHOD ejectPassengers ()V
 		COMMENT Dismounts all entities riding this entity from this entity.
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 other
 	METHOD fillCrashReportCategory (Lnet/minecraft/CrashReportCategory;)V
 		ARG 1 category
+	METHOD fillIndirectPassengers (ZLjava/util/Set;)V
+		ARG 1 checkPlayer
```
sciwhiz12: Why `checkPlayer`? From how it's used -- to skip any (server)player entities from being added to the set passed to the method -- I'd say a more appropriate name would be along the lines of `skipPlayer`.

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -139,13 +149,21 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 1 livingEntity
 	METHOD getEncodeId ()Ljava/lang/String;
 		COMMENT Returns the string that identifies this Entity's class
+	METHOD getExitPortal (Lnet/minecraft/server/level/ServerLevel;Lnet/minecraft/core/BlockPos;Z)Ljava/util/Optional;
+		ARG 1 destination
+		ARG 2 findFrom
```
sciwhiz12, line 154: I'm pondering whether this should be renamed slightly to relay the fact that this is the center position from which the exit portal finding starts.

sciwhiz12 (commented):
I have two small comments, otherwise the rest looks good to me.

data/net/minecraft/world/entity/Entity.mapping
ChampionAsh5357: I would use a javadoc in that case to relay the fact.

data/net/minecraft/world/entity/Entity.mapping
tmvkrpxl0: I think it should be left unchanged, or renamed `onlyPlayer`, I think it's for checking if It should only include `ServerPlayer`, or every entities

tmvkrpxl0: I think it should be left unchanged, or renamed `onlyPlayer`, I think it's for determining if It should only include `ServerPlayer`, or every entities

data/net/minecraft/world/entity/Entity.mapping
tmvkrpxl0: I have added javadoc for it

data/net/minecraft/world/entity/Entity.mapping
sciwhiz12: Hmmm, it seems I've misread the code. If we look at the loop as seen via Enigma:
```java
if (!onlyPlayers || ServerPlayer.class.isAssignableFrom(passenger.getClass())) {
    fillTo.add(passenger);
}
passenger.fillIndirectPassengers(skipPlayer, fillTo)
```
So the passenger entity is added to the set if either `onlyPlayers` is `false` or the passenger entity is an instance of `ServerPlayer`. So yeah, disregard my comment, but I would request that, as you suggest, it be renamed to `onlyPlayers` instead of `checkPlayer` (as the new name is more intuitively understandable as adding only players to the set).

data/net/minecraft/world/entity/Entity.mapping
sciwhiz12: Please use a javadoc for the parameter instead of putting the information in the method javadoc.

tmvkrpxl0: I moved javadoc to parameter

data/net/minecraft/world/entity/Entity.mapping
tmvkrpxl0: I have changed parameter name to `onlyPlayers`

sciwhiz12 (approved):

Harleyoc1 (approved):

PR #72, opened 2021-10-04
Mapped some unmapped parameters in n/m/w/e/Entity

tmvkrpxl0:
I hope I mapped all unmapped ones there

### DIFF
--- a/data/net/minecraft/world/entity/Entity.mapping
+++ b/data/net/minecraft/world/entity/Entity.mapping
@@ -616,5 +616,5 @@ CLASS net/minecraft/world/entity/Entity
 			ARG 4 save
 	CLASS MovementEmission
 		METHOD <init> (Ljava/lang/String;IZZ)V
-			ARG 3 emitsSounds
-			ARG 4 emitsEvents
+			ARG 3 sounds
+			ARG 4 events
--- a/data/net/minecraft/world/entity/Entity.mapping
+++ b/data/net/minecraft/world/entity/Entity.mapping
@@ -612,8 +612,8 @@ CLASS net/minecraft/world/entity/Entity
 			ARG 6 z
 	CLASS RemovalReason
 		METHOD <init> (Ljava/lang/String;IZZ)V
-			ARG 3 shouldDestroy
-			ARG 4 shouldSave
+			ARG 3 destroy
+			ARG 4 save
 	CLASS MovementEmission
 		METHOD <init> (Ljava/lang/String;IZZ)V
 			ARG 3 emitsSounds
--- a/data/net/minecraft/world/entity/Entity.mapping
+++ b/data/net/minecraft/world/entity/Entity.mapping
@@ -469,7 +469,7 @@ CLASS net/minecraft/world/entity/Entity
 	METHOD setDeltaMovement (Lnet/minecraft/world/phys/Vec3;)V
 		ARG 1 motion
 	METHOD setGlowingTag (Z)V
-		ARG 1 glowing
+		ARG 1 hasGlowingTag
 	METHOD setId (I)V
 		ARG 1 id
 	METHOD setInvisible (Z)V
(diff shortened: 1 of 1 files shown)


data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -581,3 +604,17 @@ CLASS net/minecraft/world/entity/Entity
...
+	CLASS MoveFunction
+		METHOD accept (Lnet/minecraft/world/entity/Entity;DDD)V
+			ARG 1 entity
+			ARG 2 x
+			ARG 4 y
+			ARG 6 z
+	CLASS RemovalReason
+		METHOD <init> (Ljava/lang/String;IZZ)V
+			ARG 3 shouldDestroy
+			ARG 4 shouldSave
```
Harleyoc1: (field names)
```suggestion
			ARG 3 destroy
			ARG 4 save
```

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -581,3 +604,17 @@ CLASS net/minecraft/world/entity/Entity
...
+	CLASS MovementEmission
+		METHOD <init> (Ljava/lang/String;IZZ)V
+			ARG 3 emitsSounds
+			ARG 4 emitsEvents
```
Harleyoc1: (field names)
```suggestion
			ARG 3 sounds
			ARG 4 events
```

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -457,13 +468,17 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 5 z
 	METHOD setDeltaMovement (Lnet/minecraft/world/phys/Vec3;)V
 		ARG 1 motion
+	METHOD setGlowingTag (Z)V
+		ARG 1 glowing
```
Harleyoc1: (field name)
```suggestion
		ARG 1 hasGlowingTag
```

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -142,6 +149,8 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 1 livingEntity
 	METHOD getEncodeId ()Ljava/lang/String;
 		COMMENT Returns the string that identifies this Entity's class
+	METHOD getExitPortal (Lnet/minecraft/server/level/ServerLevel;Lnet/minecraft/core/BlockPos;Z)Ljava/util/Optional;
+		ARG 3 isToNether
```
Harleyoc1: Could add mappings for the `level` and `pos` params.

Harleyoc1 (changes requested):

data/net/minecraft/world/entity/Entity.mapping
tmvkrpxl0: Done, but it seems like I missed some, ones that are red in enigma should be ones that I missed

tmvkrpxl0: I ignored one that are not "float1" or "integer1" but sometimes named parameters had those names. I should recheck every parameter

tmvkrpxl0:
Ok, I think I remapped everything, as I don't see red parameters in enigma

data/net/minecraft/world/entity/Entity.mapping
```diff
@@ -475,11 +554,15 @@ CLASS net/minecraft/world/entity/Entity
 		ARG 1 x
 		ARG 3 y
 		ARG 5 z
+	METHOD setPacketCoordinates (Lnet/minecraft/world/phys/Vec3;)V
+		ARG 1 coordinates
```
Harleyoc1: Again, try to keep the names consistent with the official fields/methods (match field names where possible). 
```suggestion
		ARG 1 packetCoordinates
```

Harleyoc1 (changes requested):
One other minor comment.

Harleyoc1 (approved):

sciwhiz12 (approved):

PR #75, opened 2021-10-07
some packages in net.minecraft.client.gui

MelanX:
net.minecraft.client.gui.chat.*
net.minecraft.client.gui.font.*
net.minecraft.client.gui.narration.*


sciwhiz12 (dismissed):

data/net/minecraft/client/gui/font/glyphs/BakedGlyph.mapping
```diff
@@ -1,4 +1,16 @@
 CLASS net/minecraft/client/gui/font/glyphs/BakedGlyph
+	METHOD <init> (Lnet/minecraft/client/renderer/RenderType;Lnet/minecraft/client/renderer/RenderType;Lnet/minecraft/client/renderer/RenderType;FFFFFFFF)V
+		ARG 1 normalType
+		ARG 2 seeThroughType
+		ARG 3 polygonOffsetType
+		ARG 4 u0
+		ARG 5 u1
+		ARG 6 v0
+		ARG 7 v1
```
marchermans: This needs proper naming
What is u0 and u1?
Why are they indexed.....

data/net/minecraft/client/gui/font/glyphs/BakedGlyph.mapping
```diff
@@ -15,3 +27,16 @@ CLASS net/minecraft/client/gui/font/glyphs/BakedGlyph
...
 		ARG 3 buffer
 		ARG 4 packedLight
+	METHOD renderType (Lnet/minecraft/client/gui/Font$DisplayMode;)Lnet/minecraft/client/renderer/RenderType;
+		ARG 1 displayMode
+	CLASS Effect
+		METHOD <init> (FFFFFFFFF)V
+			ARG 1 x0
+			ARG 2 y0
+			ARG 3 x1
+			ARG 4 y1
```
marchermans, line 41: Same here. Don't use index arguments if you can properly describe what they do

marchermans (dismissed):
Some minor things with the bakedglyph using index arguments instead of proper names.
But this looks okey for the rest.

data/net/minecraft/client/gui/font/glyphs/BakedGlyph.mapping
sciwhiz12: This is because these parameters are directly assigned to fields, and according to the [convention on parameter names](https://github.com/ParchmentMC/Parchment/wiki/Mapping-Conventions#parameter-names-conventions), parameters which are directly assigned to fields should be named the same.
So we either break that convention here (which I suggest against), or add clarifying javadocs as to what each parameter is.

MelanX: I'm too unsure what the parameters are exactly for, sorry.

Harleyoc1: They are sprite sheet x/y coordinates for the glyph texture. i.e. u0 and u1 are effectively min and max x coords for the glyph, etc.

data/net/minecraft/client/gui/chat/NarratorChatListener.mapping
```diff
@@ -6,5 +6,7 @@ CLASS net/minecraft/client/gui/chat/NarratorChatListener
 		ARG 3 sender
 	METHOD sayNow (Ljava/lang/String;)V
 		ARG 1 msg
```
Harleyoc1: Should be changed to `message`, especially since you mapped its overload to `message`.

data/net/minecraft/client/gui/narration/NarrationElementOutput.mapping
```diff
@@ -0,0 +1,4 @@
+CLASS net/minecraft/client/gui/narration/NarrationElementOutput
+	METHOD add (Lnet/minecraft/client/gui/narration/NarratedElementType;Lnet/minecraft/client/gui/narration/NarrationThunk;)V
+		ARG 1 type
+		ARG 2 contents
```
Harleyoc1, line 7: Overloads could be mapped (with the same names).

Harleyoc1 (changes requested):
Also want to re-iterate Sciwhiz's suggestion to add clarifying Javadoc to the `BakedGlyph` min/max sprite sheet coords.

data/net/minecraft/client/gui/font/glyphs/BakedGlyph.mapping
```diff
@@ -4,9 +4,13 @@ CLASS net/minecraft/client/gui/font/glyphs/BakedGlyph
 		ARG 2 seeThroughType
 		ARG 3 polygonOffsetType
 		ARG 4 u0
+			COMMENT minX position of the glyph texture
 		ARG 5 u1
+			COMMENT maxX position of the glyph texture
 		ARG 6 v0
+			COMMENT minY position of the glyph texture
 		ARG 7 v1
+			COMMENT maxY position of the glyph texture
```
Harleyoc1: Seems strange to use `minX`, `maxX`, etc if they don't refer to variable names. I'd suggest rewording all of these to `min/max x/y coordinate` (`of the glyph texture` is also redundant).

Harleyoc1 (changes requested):

Harleyoc1 (approved):

alcatrazEscapee (approved):

PR #50, opened 2021-07-25
Add / Update parameter mappings for net.minecraft.client.gui.components.*

WithoutAName25:

PR #76, opened 2021-10-07
additional work on net.minecraft.client.gui.components.*

MelanX:
Tried to look at #50 to not change stuff it added.

### DIFF
--- a/data/net/minecraft/client/gui/components/AbstractSelectionList.mapping
+++ b/data/net/minecraft/client/gui/components/AbstractSelectionList.mapping
@@ -65,7 +65,7 @@ CLASS net/minecraft/client/gui/components/AbstractSelectionList
 	METHOD removeEntry (Lnet/minecraft/client/gui/components/AbstractSelectionList$Entry;)Z
 		ARG 1 entry
 	METHOD render (Lcom/mojang/blaze3d/vertex/PoseStack;IIF)V
-		ARG 1 matrixStack
+		ARG 1 poseStack
 		ARG 2 mouseX
 		ARG 3 mouseY
 		ARG 4 partialTick
@@ -81,7 +81,7 @@ CLASS net/minecraft/client/gui/components/AbstractSelectionList
 		ARG 3 y
 		ARG 4 tessellator
 	METHOD renderList (Lcom/mojang/blaze3d/vertex/PoseStack;IIIIF)V
-		ARG 1 matrixStack
+		ARG 1 poseStack
 		ARG 2 x
 		ARG 3 y
 		ARG 4 mouseX
--- a/data/net/minecraft/client/gui/components/AbstractSelectionList.mapping
+++ b/data/net/minecraft/client/gui/components/AbstractSelectionList.mapping
@@ -8,8 +8,13 @@ CLASS net/minecraft/client/gui/components/AbstractSelectionList
 		ARG 6 itemHeight
 	METHOD addEntry (Lnet/minecraft/client/gui/components/AbstractSelectionList$Entry;)I
 		ARG 1 entry
+	METHOD bindEntryToSelf (Lnet/minecraft/client/gui/components/AbstractSelectionList$Entry;)V
+		ARG 1 entry
 	METHOD centerScrollOn (Lnet/minecraft/client/gui/components/AbstractSelectionList$Entry;)V
 		ARG 1 entry
+	METHOD clickedHeader (II)V
+		ARG 1 x
+		ARG 2 y
 	METHOD ensureVisible (Lnet/minecraft/client/gui/components/AbstractSelectionList$Entry;)V
 		ARG 1 entry
 	METHOD getEntry (I)Lnet/minecraft/client/gui/components/AbstractSelectionList$Entry;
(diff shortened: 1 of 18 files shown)


sciwhiz12 (dismissed):

data/net/minecraft/client/gui/components/AbstractSelectionList.mapping
```diff
@@ -85,20 +92,20 @@ CLASS net/minecraft/client/gui/components/AbstractSelectionList
 	METHOD scroll (I)V
 		ARG 1 scroll
 	METHOD setLeftPos (I)V
-		ARG 1 left
+		ARG 1 x0
```
marchermans, line 95: Don't use indexed arguments. Name them properly

data/net/minecraft/client/gui/components/AbstractSelectionList.mapping
```diff
@@ -8,8 +8,13 @@ CLASS net/minecraft/client/gui/components/AbstractSelectionList
 		ARG 6 itemHeight
 	METHOD addEntry (Lnet/minecraft/client/gui/components/AbstractSelectionList$Entry;)I
 		ARG 1 entry
+	METHOD bindEntryToSelf (Lnet/minecraft/client/gui/components/AbstractSelectionList$Entry;)V
+		ARG 1 entry
 	METHOD centerScrollOn (Lnet/minecraft/client/gui/components/AbstractSelectionList$Entry;)V
 		ARG 1 entry
+	METHOD clickedHeader (II)V
+		ARG 1 x
+		ARG 2 y
```
marchermans: If this is the mouse position, then the word "mouse" needs to be prefixed, see other locations in this file for examples.

data/net/minecraft/client/gui/components/CommandSuggestions.mapping
```diff
@@ -47,6 +47,8 @@ CLASS net/minecraft/client/gui/components/CommandSuggestions
 	METHOD sortSuggestions (Lcom/mojang/brigadier/suggestion/Suggestions;)Ljava/util/List;
 		ARG 1 suggestions
 	CLASS SuggestionsList
+		METHOD <init> (Lnet/minecraft/client/gui/components/CommandSuggestions;IIILjava/util/List;Z)V
+			ARG 5 suggestionList
```
marchermans, line 54: Any idea what the other arguments are? Or are they mapped from super types?

marchermans (dismissed):
Some minor nitpickings and questions. Looks good otherwise though

data/net/minecraft/client/gui/components/AbstractSelectionList.mapping
sciwhiz12: This is because these parameters are directly assigned to fields, and according to the [convention on parameter names](https://github.com/ParchmentMC/Parchment/wiki/Mapping-Conventions#parameter-names-conventions), parameters which are directly assigned to fields should be named the same.
So we either break that convention here (which I suggest against), or add clarifying javadocs as to what each parameter is.

data/net/minecraft/client/gui/components/CommandSuggestions.mapping
MelanX: True, looking a bit more detailed on it the other arguments are easy.

data/net/minecraft/client/gui/components/AbstractSelectionList.mapping
MelanX: I'm too unsure what the parameters are exactly for, sorry.

Harleyoc1: It's just resetting the x-coordinates based on a given position relative to the left side of the screen. Whilst it is assigned to the field `x0` it is also used to calculate `x1` so `left` or `x` is an acceptable name IMO.

data/net/minecraft/client/gui/components/spectator/SpectatorGui.mapping
```diff
@@ -9,6 +9,7 @@ CLASS net/minecraft/client/gui/components/spectator/SpectatorGui
 		ARG 1 menu
 	METHOD renderHotbar (Lcom/mojang/blaze3d/vertex/PoseStack;F)V
 		ARG 1 poseStack
+		ARG 2 partialTicks
```
Harleyoc1: Might want to do a full search and replace on this PR too. 
```suggestion
		ARG 2 partialTick
```

data/net/minecraft/client/gui/components/toasts/Toast.mapping
```diff
@@ -2,6 +2,7 @@ CLASS net/minecraft/client/gui/components/toasts/Toast
 	METHOD render (Lcom/mojang/blaze3d/vertex/PoseStack;Lnet/minecraft/client/gui/components/toasts/ToastComponent;J)Lnet/minecraft/client/gui/components/toasts/Toast$Visibility;
 		ARG 1 poseStack
 		ARG 2 toastComponent
+		ARG 3 lastChanged
```
Harleyoc1: I would recommend also adding Javadoc to clarify that it's in milliseconds (or call it `millisSinceLastVisible`). 
```suggestion
		ARG 3 timeSinceLastVisible
```

Harleyoc1 (changes requested):

Harleyoc1 (approved):

sciwhiz12 (approved):

PR #93, opened 2021-10-20
Mapped some parameters in BlockGetter

tmvkrpxl0:
`belowShapeSupplier` is guessed parameter name. All use cases for this was providing collision shape of block below another shape that is provided in first argument. I'm not sure what to name this
`traverseContext` and `traversePos` is repeatedly provided to `test` BiFunction in while loop in `traverseBlock`. `traversePos` is temporary position in this context. Since those 2 are being used when traversing block, I think this shouldn't just be `pos` or `context` because of that.
I'm not sure what to name `tester` and `onFail`. When traversing block, `tester` is repeatedly used to determine to see if it hit block. and `onFail` is executed when it misses.


data/net/minecraft/world/level/BlockGetter.mapping
```diff
@@ -1,5 +1,7 @@
 CLASS net/minecraft/world/level/BlockGetter
 	METHOD clip (Lnet/minecraft/world/level/ClipContext;)Lnet/minecraft/world/phys/BlockHitResult;
+		COMMENT Checks if there's block between {@code from} and {@code to} of context.
+		COMMENT This uses Collision shape of provided block.
```
alcatrazEscapee:
```suggestion
		COMMENT This uses the collision shape of provided block.
```

alcatrazEscapee (changes requested):

alcatrazEscapee (approved):

sciwhiz12 (approved):

PR #101, opened 2022-01-10
[1.18] A few random mappings made with scribe

MelanX:
While coding, I saw a few mappings. Here is what I changed.

### DIFF
--- a/data/net/minecraft/data/models/BlockModelGenerators.mapping
+++ b/data/net/minecraft/data/models/BlockModelGenerators.mapping
@@ -326,6 +326,8 @@ CLASS net/minecraft/data/models/BlockModelGenerators
 	CLASS WoodProvider
 		METHOD <init> (Lnet/minecraft/data/models/BlockModelGenerators;Lnet/minecraft/data/models/model/TextureMapping;)V
 			ARG 2 logMapping
+		METHOD <init> (Lnet/minecraft/data/models/model/TextureMapping;)V
+			ARG 1 textureMapping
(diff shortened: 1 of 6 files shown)


data/net/minecraft/data/models/BlockModelGenerators.mapping
```diff
@@ -326,6 +326,8 @@ CLASS net/minecraft/data/models/BlockModelGenerators
 	CLASS WoodProvider
 		METHOD <init> (Lnet/minecraft/data/models/BlockModelGenerators;Lnet/minecraft/data/models/model/TextureMapping;)V
 			ARG 2 logMapping
+		METHOD <init> (Lnet/minecraft/data/models/model/TextureMapping;)V
+			ARG 1 textureMapping
```
sciwhiz12: Hmm, this looks like an edge case with inner classes (non-`static` nested classes), as the constructor of this method has two arguments, as can be seen in the correct mapping above (the enclosing class instance, and then the regular mandated parameters). @SizableShrimp, could you please take a look at this?

In the meantime, please remove this and the other added mapping in this class.

sciwhiz12 (changes requested):

data/net/minecraft/data/models/BlockModelGenerators.mapping
sciwhiz12: This issue is now being tracked at https://github.com/ParchmentMC/Scribe/issues/5.

sciwhiz12 (approved):

SizableShrimp (approved):

PR #108, opened 2022-02-19
Some mappings in net.minecraft.server

noeppi-noeppi:


data/net/minecraft/server/players/IpBanListEntry.mapping
```diff
@@ -1,4 +1,10 @@
 CLASS net/minecraft/server/players/IpBanListEntry
+	METHOD <init> (Lcom/google/gson/JsonObject;)V
+		ARG 1 entryData
+	METHOD <init> (Ljava/lang/String;)V
+		ARG 1 ip
+	METHOD <init> (Ljava/lang/String;Ljava/util/Date;Ljava/lang/String;Ljava/util/Date;Ljava/lang/String;)V
+		ARG 1 ip
```
sciwhiz12, line 7: Why wasn't the other parameters of this constructor named, given that they are available from looking at the superclass constructor which it calls?

data/net/minecraft/server/players/StoredUserList.mapping
```diff
@@ -13,5 +15,7 @@ CLASS net/minecraft/server/players/StoredUserList
 		ARG 1 obj
 	METHOD remove (Ljava/lang/Object;)V
 		ARG 1 entry
+	METHOD remove (Lnet/minecraft/server/players/StoredUserEntry;)V
+		ARG 1 user
```
sciwhiz12: The naming of the parameters of these twomethods should be interchanged; the first one removes a user, while the second one accepts an entry and removes the user for that entry.

sciwhiz12 (changes requested):
Just have two comments of note, the rest looks fine to me.

sciwhiz12 (approved):

Lanse505 (approved):

PR #109, opened 2022-02-19
Improve mappings in net.minecraft.world.level.chunk.storage

noeppi-noeppi:


data/net/minecraft/world/level/chunk/storage/SectionStorage.mapping
```diff
@@ -7,5 +7,37 @@ CLASS net/minecraft/world/level/chunk/storage/SectionStorage
 		ARG 5 type
 		ARG 6 sync
 		ARG 7 levelHeightAccessor
+	METHOD flush (Lnet/minecraft/world/level/ChunkPos;)V
+		ARG 1 chunkPos
+	METHOD get (J)Ljava/util/Optional;
+		ARG 1 packedSectionId
```
sciwhiz12: For this and equally-named parameters, perhaps it would be better named `sectionKey`, referencing the static `getKey` method in the same method?

sciwhiz12 (changes requested):
Just one small comment; the rest looks fine to me.

sciwhiz12 (approved):

Lanse505 (approved):

PR #119, opened 2022-05-15
Update old javadoc in Item.isFoil

Technici4n:
image

### DIFF
--- a/data/net/minecraft/world/item/Item.mapping
+++ b/data/net/minecraft/world/item/Item.mapping
@@ -102,7 +102,7 @@ CLASS net/minecraft/world/item/Item
 		COMMENT Checks isDamagable and if it cannot be stacked
 		ARG 1 stack
 	METHOD isFoil (Lnet/minecraft/world/item/ItemStack;)Z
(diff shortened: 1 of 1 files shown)


data/net/minecraft/world/item/Item.mapping
```diff
@@ -102,7 +102,7 @@ CLASS net/minecraft/world/item/Item
 		COMMENT Checks isDamagable and if it cannot be stacked
 		ARG 1 stack
 	METHOD isFoil (Lnet/minecraft/world/item/ItemStack;)Z
-		COMMENT Returns true if this item has an enchantment glint. By default, this returns <code>stack.isItemEnchanted()</code>, but other items can override it (for instance, written books always return true).
+		COMMENT Returns true if this item has an enchantment glint. By default, this returns <code>stack.isEnchanted()</code>, but other items can override it (for instance, written books always return true).
```
sciwhiz12: Since it's already being modified, the HTML tags should be replaced with the javadoc `{@code}` tag.
```suggestion
		COMMENT Returns true if this item has an enchantment glint. By default, this returns {@code stack.isEnchanted()}, but other items can override it (for instance, written books always return true).
```

sciwhiz12 (changes requested):

sciwhiz12 (approved):

AterAnimAvis (approved):

PR #124, opened 2022-06-19
[1.18.x] Cherrypick 1.16.x PRs

SizableShrimp:
Needs a thorough review. Some standards have changed since these 1.16 PRs which may need work. I tried to keep as much data as possible. `sanitizeData` only dropped one method in Zoglin which was just due to it changing from static to an instance method, was easy to fix manually.


alcatrazEscapee (approved):
Alright, seeing as these are al cherry picks from 1.16.x I'm not reviewing these carefully as per standards, since they've already been accepted once. I did look through for any irregularities and it seems you cherry picked successfully, and there's not many javadocs here that may have gone out of date / become wrong between the two versions (nor did these commits touch areas that I'm aware changed drastically across those two versions).

So, LGTM!

sciwhiz12 (approved):

PR #127, opened 2022-06-23
Added various mappings

PiTheGuy:


PiTheGuy:
Found validation errors, need to fix before this is ready for review

PiTheGuy:
This pull request is ready,

alcatrazEscapee:
So just to clarify one point: the validation error with methods like `compareTo` is due to a quirk with how Enigma ends up mapping methods - it maps the synthetic bridge method rather than the user visible method. Meaning you mapped `compareTo(Object)` instead of `<T> compareTo(T)`. The fix for this is manually changing the signature, in the `.mapping` file, if you so desire.

@sciwhiz12 wrote a whole article on synthetic bouncers in Java and how they apply here if he wants to plug it :)

PiTheGuy:
Ok, I'll fix that real quick

PiTheGuy:
What do I change the signature to?

alcatrazEscapee:
`compareTo(T t)` is from `Comparable<T>`, which will be erased to whatever the type bound on `Comparable`  is. Then use the fully qualified descriptor, so if it implements `Comparable<Foo>`, you want to replace the `Object` with `Foo`

That said, if you don't want to screw around with the raw .mapping files, it's entirely fine to just omit them.

data/net/minecraft/server/commands/DefaultGameModeCommands.mapping
```diff
@@ -2,6 +2,6 @@ CLASS net/minecraft/server/commands/DefaultGameModeCommands
 	METHOD register (Lcom/mojang/brigadier/CommandDispatcher;)V
 		ARG 0 dispatcher
 	METHOD setMode (Lnet/minecraft/commands/CommandSourceStack;Lnet/minecraft/world/level/GameType;)I
-		COMMENT Set Gametype of player who ran the command
+		COMMENT Set Gametype of the player who ran the command
```
alcatrazEscapee:
```suggestion
		COMMENT Sets the {@link net.minecraft.world.level.GameType} of the player who ran the command.
```

alcatrazEscapee (changes requested):
Each time I went to comment that a parameter name was not great, I went and looked it up, and it was because it was matching an existing mojmap name. Every *single* time.

The vast majority of this PR is datafixer and schema names, of which there's about two reviewable changes spread over some two hundred files it seems.

Just one comment, as if we're changing this javadoc we might as well properly fix it.

alcatrazEscapee:
Two quick comments: please don't keep adding just everything-and-the-kitchen-sink to this PR, as larger PRs make it harder to review them.

Additionally, please don't continue to add mappings after you've stated a PR is ready for review, because as again, it makes it harder to review.

PiTheGuy:
Noted, I'll make a new PR from now on

alcatrazEscapee (approved):

PR #136, opened 2022-06-30
Various mappings for net/minecraft/world/level/block

PiTheGuy:


data/net/minecraft/world/level/block/AbstractCandleBlock.mapping
```diff
@@ -4,6 +4,11 @@ CLASS net/minecraft/world/level/block/AbstractCandleBlock
 		ARG 2 random
 	METHOD canBeLit (Lnet/minecraft/world/level/block/state/BlockState;)Z
 		ARG 1 state
+	METHOD extinguish (Lnet/minecraft/world/entity/player/Player;Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/LevelAccessor;Lnet/minecraft/core/BlockPos;)V
+		ARG 0 player
+		ARG 1 state
+		ARG 2 levelAccessor
```
alcatrazEscapee:
```suggestion
		ARG 2 level
```

data/net/minecraft/world/level/block/MultifaceBlock.mapping
```diff
@@ -1,12 +1,37 @@
...
 		ARG 1 face
 	METHOD lambda$getStateForPlacement$1 (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/Direction;)Lnet/minecraft/world/level/block/state/BlockState;
 		ARG 4 lookingDirection
 	METHOD lambda$static$0 (Ljava/util/EnumMap;)V
 		ARG 0 map
+	METHOD mapDirections (Lnet/minecraft/world/level/block/state/BlockState;Ljava/util/function/Function;)Lnet/minecraft/world/level/block/state/BlockState;
+		ARG 1 state
+	METHOD spreadFromFaceTowardDirection (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/LevelAccessor;Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/Direction;Lnet/minecraft/core/Direction;Z)Z
+		ARG 1 state
+		ARG 2 levelAccessor
```
alcatrazEscapee:
```suggestion
		ARG 2 level
```

data/net/minecraft/world/level/block/MultifaceBlock.mapping
```diff
@@ -1,12 +1,37 @@
...
+		ARG 3 pos
+		ARG 4 face
+		ARG 5 direction
+	METHOD spreadFromFaceTowardRandomDirection (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/LevelAccessor;Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/Direction;Ljava/util/Random;Z)Z
+		ARG 1 state
+		ARG 2 levelAccessor
```
alcatrazEscapee:
```suggestion
		ARG 2 level
```

data/net/minecraft/world/level/block/VineBlock.mapping
```diff
@@ -30,6 +30,10 @@ CLASS net/minecraft/world/level/block/VineBlock
 		ARG 4 context
 	METHOD getStateForPlacement (Lnet/minecraft/world/item/context/BlockPlaceContext;)Lnet/minecraft/world/level/block/state/BlockState;
 		ARG 1 context
+	METHOD getUpdatedState (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/BlockGetter;Lnet/minecraft/core/BlockPos;)Lnet/minecraft/world/level/block/state/BlockState;
+		ARG 1 state
+		ARG 2 blockGetter
```
alcatrazEscapee:
```suggestion
		ARG 2 level
```

data/net/minecraft/world/level/block/entity/BaseContainerBlockEntity.mapping
```diff
@@ -4,6 +4,7 @@ CLASS net/minecraft/world/level/block/entity/BaseContainerBlockEntity
 	METHOD canUnlock (Lnet/minecraft/world/entity/player/Player;Lnet/minecraft/world/LockCode;Lnet/minecraft/network/chat/Component;)Z
 		ARG 0 player
 		ARG 1 code
+		ARG 2 containerTitle
```
alcatrazEscapee:
```suggestion
		ARG 2 displayName
```
looking at usages, that seems to fit better.

data/net/minecraft/world/level/block/entity/BellBlockEntity.mapping
```diff
@@ -3,11 +3,26 @@ CLASS net/minecraft/world/level/block/entity/BellBlockEntity
 		COMMENT Warmup counter until raiders are revealed by glow and particle
 	FIELD ticks I
 		COMMENT How long the bell has been ringing
+	METHOD <init> (Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/level/block/state/BlockState;)V
+		ARG 1 worldPosition
```
alcatrazEscapee:
```suggestion
		ARG 1 pos
```
Despite the fact that in `BlockEntity` (and seemingly *only* in `BlockEntity`), the field this is eventually assigned to is called `worldPosition`, I'd advocate for keeping consistency with other `pos` parameters in this call path.

Some other notes:
- We currently map `BlockEntity.<init>` as `worldPosition` [here](https://github.com/ParchmentMC/Parchment/blob/3df8090b42497d2779dee6ff0850f4ae8322193d/data/net/minecraft/world/level/block/entity/BlockEntity.mapping#L4).
- The accessor for `BlockEntity.worldPosition` is called `getBlockPos()`, not `getWorldPosition()`. Meaning from the public API (constructor and accessor), you never see the name `worldPosition`.

cc @ParchmentMC/mapping-standards (because I want a second opinion on this one)

data/net/minecraft/world/level/block/piston/MovingPistonBlock.mapping
```diff
@@ -33,6 +33,13 @@ CLASS net/minecraft/world/level/block/piston/MovingPistonBlock
 	METHOD mirror (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/block/Mirror;)Lnet/minecraft/world/level/block/state/BlockState;
 		ARG 1 state
 		ARG 2 mirror
+	METHOD newMovingBlockEntity (Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/core/Direction;ZZ)Lnet/minecraft/world/level/block/entity/BlockEntity;
+		ARG 0 worldPosition
```
alcatrazEscapee:
```suggestion
		ARG 0 pos
```
Same note about `worldPosition` as above.

alcatrazEscapee (changes requested):

data/net/minecraft/world/level/block/VineBlock.mapping
PiTheGuy: I kind of wondered about this. Will change in next commit.

data/net/minecraft/world/level/block/MultifaceBlock.mapping
PiTheGuy: Can you explain your reasoning on why this is better? I'm open to changing it, I just want to know why so I can learn for the future.

alcatrazEscapee: Most of the usages of `Level` interfaces don't really have a specific reason for being typed as one of the interfaces as opposed to a different one - or at least, ones that we can clearly see. In most cases, these all refer to the same object (or, one of a few, i.e. the `level` in world gen is generally not the same as the `level` in server code vs. a `level` in client code.

In the absence of needing this parameter to be a specific interface, and since `level` is such a common occurrence throughout the code, these interfaces should generally be shortened just to `level`.

As an example where this *isn't* the case, look at `ChunkAccess.levelHeightAccessor`. This is explicitly typed very vaguely, as chunks need height information, but no guarantee is made that this is a normal `level` for generic use - and we can see that i.e. in `ProtoChunk`, this may contain a reference to the `ServerLevel`, which is a very big pitfall if you treat that as the currently generating level during world gen, because you'll create a circular chunk loading deadlock.

data/net/minecraft/world/level/block/entity/BellBlockEntity.mapping
sciwhiz12: I'd be fine with mapping this specific parameter in the `BlockEntity` and subclasses' constructors as `pos`, for consistency.

PiTheGuy: I think I already committed the suggestion earlier.

alcatrazEscapee (approved):

PR #150, opened 2022-07-10
Refactor worldPosition -> pos, and cleanup some misc. javadocs.

alcatrazEscapee:
Continuing from https://github.com/ParchmentMC/Parchment/pull/136#discussion_r912521702, it was decided to switch `worldPosition` -> `pos`, for the reasons outlined in said comment (consistency with public accessor `getBlockPos` over protected field `worldPosition`, `worldPosition` itself being a weird name only used by block entities, and consistency with other `BlockPos` arguments). This simply makes that change across all existing mapped block entities.

While I was going through these, I edited a few javadocs, expanding, fixing formatting (sentences), or just general cleanup.

### DIFF
--- a/data/net/minecraft/world/level/block/entity/BannerBlockEntity.mapping
+++ b/data/net/minecraft/world/level/block/entity/BannerBlockEntity.mapping
@@ -7,7 +7,7 @@ CLASS net/minecraft/world/level/block/entity/BannerBlockEntity
 		ARG 1 worldPosition
 		ARG 2 blockState
 	METHOD <init> (Lnet/minecraft/core/BlockPos;Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/item/DyeColor;)V
-		ARG 1 worldPosition
+		ARG 1 pos
 		ARG 2 blockState
 		ARG 3 baseColor
 	METHOD createPatterns (Lnet/minecraft/world/item/DyeColor;Lnet/minecraft/nbt/ListTag;)Ljava/util/List;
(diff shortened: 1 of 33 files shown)


data/net/minecraft/world/level/block/entity/HopperBlockEntity.mapping
```diff
@@ -39,21 +39,31 @@ CLASS net/minecraft/world/level/block/entity/HopperBlockEntity
...
 	METHOD isEmptyContainer (Lnet/minecraft/world/Container;Lnet/minecraft/core/Direction;)Z
-		COMMENT @return whether the given Container is empty from the given face
+		COMMENT @return whether the given {@link Container} is empty from the given face
 		ARG 0 container
 		ARG 1 direction
 	METHOD isFullContainer (Lnet/minecraft/world/Container;Lnet/minecraft/core/Direction;)Z
-		COMMENT @return false if the container has any room to place items in
+		COMMENT @return false if the {@link Container} has any room to place items in
 		ARG 0 container
 		ARG 1 direction
```
Harleyoc1: I would advise doing `{@code container}` here since it's a reference to a parameter. 
To be consistent with your other changes I would also change `false` to `{@code false}`.

Harleyoc1 (dismissed):

data/net/minecraft/world/level/block/entity/BannerBlockEntity.mapping
```diff
@@ -7,7 +7,7 @@ CLASS net/minecraft/world/level/block/entity/BannerBlockEntity
 		ARG 1 worldPosition
```
sciwhiz12: Missed this one.
```suggestion
		ARG 1 pos
```

sciwhiz12 (changes requested):
`net\minecraft\world\level\block\piston\PistonMovingBlockEntity` also needs to be corrected from `worldPosition` parameter names to `pos`.

sciwhiz12 (approved):

PR #164, opened 2022-08-12
Update the game version in the readme

PiTheGuy:


SizableShrimp:
I don't really get why this branch was made on the main repo (or why ppl like us have access to make branches cc @sciwhiz12)

PiTheGuy:
I used the edit button on the main page. I thought it would create a new branch in my fork, not the main repo. As to why it let me make the branch in the first place, I don't know.

sciwhiz12:
@SizableShrimp, GitHub by default allows repository collaborators with push access to create branches on the repository, and only a branch protection rule can prevent creation of such branches. Before now, the only branch protection rule we had were for the `versions/*` branches.

Correspondingly, I've created a new branch protection rule which prevents anyone except repository administrators from making branches here, so that should prevent this from reoccurring. If it becomes a problem somehow, we can revisit this branch protection rule at a later date.

sciwhiz12 (approved):

PR #173, opened 2022-09-01
Fix: Incorrect argument name in net.minecraft.world.inventory.AbstractContainerMenu#clicked & doClick

MarbleGateKeeper:
I only fixed it on 1.18.x, and there is also same mistake on other branches.

### DIFF
--- a/data/net/minecraft/world/inventory/AbstractContainerMenu.mapping
+++ b/data/net/minecraft/world/inventory/AbstractContainerMenu.mapping
@@ -39,13 +39,13 @@ CLASS net/minecraft/world/inventory/AbstractContainerMenu
 		ARG 1 player
 		ARG 2 id
 	METHOD clicked (IILnet/minecraft/world/inventory/ClickType;Lnet/minecraft/world/entity/player/Player;)V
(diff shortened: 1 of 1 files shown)


data/net/minecraft/world/inventory/AbstractContainerMenu.mapping
```diff
@@ -39,13 +39,13 @@ CLASS net/minecraft/world/inventory/AbstractContainerMenu
 		ARG 1 player
 		ARG 2 id
 	METHOD clicked (IILnet/minecraft/world/inventory/ClickType;Lnet/minecraft/world/entity/player/Player;)V
-		ARG 1 mouseX
-		ARG 2 mouseY
+		ARG 1 slotId
+		ARG 2 dragType
```
sciwhiz12: The crash report detail in the method seems to imply that this (and the same parameter in `doClick`) is for the mouse button, not the 'drag type' (which is also unclear as a name, in what it refers to).

sciwhiz12 (changes requested):

sciwhiz12 (approved):

PR #183, opened 2022-10-31
Two backports and some net/minecraft/server mappings

FlowArg:
have a good day :^)

### DIFF
--- a/data/net/minecraft/server/packs/resources/SimpleJsonResourceReloadListener.mapping
+++ b/data/net/minecraft/server/packs/resources/SimpleJsonResourceReloadListener.mapping
@@ -1,6 +1,6 @@
 CLASS net/minecraft/server/packs/resources/SimpleJsonResourceReloadListener
 	METHOD <init> (Lcom/google/gson/Gson;Ljava/lang/String;)V
-		ARG 1 gson3
+		ARG 1 gson
 		ARG 2 directory
 	METHOD prepare (Lnet/minecraft/server/packs/resources/ResourceManager;Lnet/minecraft/util/profiling/ProfilerFiller;)Ljava/util/Map;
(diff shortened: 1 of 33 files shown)


data/net/minecraft/server/packs/resources/SimpleJsonResourceReloadListener.mapping
```diff
@@ -1,4 +1,7 @@
 CLASS net/minecraft/server/packs/resources/SimpleJsonResourceReloadListener
+	METHOD <init> (Lcom/google/gson/Gson;Ljava/lang/String;)V
+		ARG 1 gson3
```
sciwhiz12:
```suggestion
		ARG 1 gson
```

data/net/minecraft/server/rcon/thread/RconClient.mapping
```diff
@@ -8,3 +12,4 @@ CLASS net/minecraft/server/rcon/thread/RconClient
 		COMMENT Sends the standard RCon 'authorization failed' response packet
 	METHOD sendCmdResponse (ILjava/lang/String;)V
 		COMMENT Splits the response message into individual packets and sends each one
+		ARG 1 exception
```
sciwhiz12: Looking at the [docs for the Source RCON Protocol](https://developer.valvesoftware.com/wiki/Source_RCON_Protocol), which is the protocol that Minecraft implements for RCON, this value is transmitted as the Packet ID. See the implementation of the method, which calls `send(int, int, String)`, where the parameters are the Packet ID, Packet Type (in this case, `0` for `SERVERDATA_RESPONSE_VALUE`) and the Body (the response of the server).

```suggestion
		ARG 1 id
```

sciwhiz12 (changes requested):

sciwhiz12 (approved):

PR #197, opened 2023-03-15
Update to 1.19.4

PiTheGuy:


PiTheGuy:
Continued in #198 because this one got messed up

PR #200, opened 2023-04-02
redstone signal javadocs and related refactors

SpaceWalkerRS:


data/net/minecraft/world/level/block/state/BlockBehaviour.mapping
```diff
@@ -189,8 +204,14 @@ CLASS net/minecraft/world/level/block/state/BlockBehaviour
 		ARG 4 id
 		ARG 5 param
 	METHOD updateIndirectNeighbourShapes (Lnet/minecraft/world/level/block/state/BlockState;Lnet/minecraft/world/level/LevelAccessor;Lnet/minecraft/core/BlockPos;II)V
-		COMMENT Performs updates on diagonal neighbors of the target position and passes in the flags.
-		COMMENT The flags are equivalent to {@link net.minecraft.world.level.Level#setBlock}.
+		COMMENT Updates the shapes of indirect neighbors of this block. This method is analogous to
+		COMMENT {@link net.minecraft.world.level.block.state.BlockBehaviour$BlockStateBase#updateNeighbourShapes}
+		COMMENT but where that method affects the 6 direct neighbors of this block, this method affects
+		COMMENT indirect the neighbors, if any.
```
sciwhiz12: Looks like a simple switch-a-roo of words.
```suggestion
		COMMENT the indirect neighbors, if any.
```

sciwhiz12 (changes requested):
Hello! Sorry for the delay in your PR getting a review; I've been a tad busy due to personal commitments and some issues in the toolchain (which were fixed today).

Looks fine to me; just have that one grammar fix. 

(Also, the fact that redstone diodes have the opposite of the instinctual facing -- their facing is aligned to ther _input_, rather than their _output_ as with all other blocks like dispensers, droppers, hoppers, etc. -- is fun _and_ confusing. 😄)

SpaceWalkerRS:
No worries, thanks for the review! There have been some changes to this code in 1.20 so look forward to another PR for that ^^

> (Also, the fact that redstone diodes have the opposite of the instinctual facing -- their facing is aligned to ther input, rather than their output as with all other blocks like dispensers, droppers, hoppers, etc. -- is fun and confusing. 😄)

Yeah that's quite baffling to be honest - I don't know if it's a remnant of old Notch code or what, but surely that's something that could/would have been fixed in 1.8, if not during The Flattening? Oh well.

sciwhiz12 (approved):

PR #201, opened 2023-06-07
Update to 1.20.1

PiTheGuy:
Also includes an enigma update to 2.3.0


build.gradle
```diff
@@ -33,7 +33,7 @@ writtenbooks {
 }
 
 compass {
-    version = '1.19.4'
+    version = '1.20.1-rc1'
```
sciwhiz12: I believe we want to target `1.20.1` now.

build.gradle
```diff
@@ -60,7 +60,7 @@ dependencies {
     remapper 'net.minecraftforge:ForgeAutoRenamingTool:0.1.24'
```
sciwhiz12: We should also update ForgeAutoRenamingTool to `1.0.2`.

sciwhiz12 (changes requested):

sciwhiz12 (approved):

PR #202, opened 2023-06-26
redstone signal javadocs and related refactors (1.20 edition)

SpaceWalkerRS:
some of the changes from #200 got lost in the update to 1.20, and on top of that, some of the code that used to be in `Level` got moved into its own interface, `SignalGetter`, so I ported the javadocs over to that

### DIFF
--- a/data/net/minecraft/world/level/SignalGetter.mapping
+++ b/data/net/minecraft/world/level/SignalGetter.mapping
@@ -1,19 +1,50 @@
 CLASS net/minecraft/world/level/SignalGetter
 	METHOD getBestNeighborSignal (Lnet/minecraft/core/BlockPos;)I
+		COMMENT Returns the highest redstone signal the given position receives from neighboring blocks.
 		ARG 1 pos
 	METHOD getControlInputSignal (Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/Direction;Z)I
+		COMMENT Returns the control signal emitted from the given position in the given direction.
+		COMMENT If {@code diodesOnly} is {@code true}, this method returns the direct signal emitted if
+		COMMENT and only if this position is occupied by a diode (i.e. a repeater or comparator).
+		COMMENT Otherwise, if this position is occupied by a redstone block, this method will return the
+		COMMENT redstone signal emitted by it. If not, this method will return the direct signal emitted
+		COMMENT from this position in the given direction.
+		COMMENT
+		COMMENT <p>
+		COMMENT NOTE: directions in redstone signal related methods are backwards, so this method
+		COMMENT checks for the signal emitted in the <i>opposite</i> direction of the one given.
 		ARG 1 pos
 		ARG 2 direction
+		ARG 3 diodesOnly
 	METHOD getDirectSignal (Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/Direction;)I
+		COMMENT Returns the direct redstone signal emitted from the given position in the given direction.
+		COMMENT
+		COMMENT <p>
+		COMMENT NOTE: directions in redstone signal related methods are backwards, so this method
+		COMMENT checks for the signal emitted in the <i>opposite</i> direction of the one given.
 		ARG 1 pos
 		ARG 2 direction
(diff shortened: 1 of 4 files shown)


data/net/minecraft/world/level/SignalGetter.mapping
```diff
@@ -1,19 +1,50 @@
 CLASS net/minecraft/world/level/SignalGetter
 	METHOD getBestNeighborSignal (Lnet/minecraft/core/BlockPos;)I
+		COMMENT Returns the highest redstone signal the given position receives from neighboring blocks.
 		ARG 1 pos
 	METHOD getControlInputSignal (Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/Direction;Z)I
+		COMMENT Returns the control signal emitted from the given position in the given direction.
+		COMMENT If {@code diodesOnly} is {@code true}, this method returns the direct signal emitted if
+		COMMENT and only if this position is occupied by a diode (i.e. a repeater or comparator).
+		COMMENT Otherwise, if this position is occupied by a redstone block, this method will return the
```
sciwhiz12: Just to clarify, when you say "redstone block", you mean the block made from 9 redstone dust which emits a signal strength of 15? Might be good to `{@linkplain}` to `Blocks#REDSTONE_BLOCK` to clarify that in the comment.

data/net/minecraft/world/level/SignalGetter.mapping
```diff
@@ -1,19 +1,50 @@
 CLASS net/minecraft/world/level/SignalGetter
 	METHOD getBestNeighborSignal (Lnet/minecraft/core/BlockPos;)I
+		COMMENT Returns the highest redstone signal the given position receives from neighboring blocks.
 		ARG 1 pos
 	METHOD getControlInputSignal (Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/Direction;Z)I
+		COMMENT Returns the control signal emitted from the given position in the given direction.
```
sciwhiz12, line 6: what's a `control signal`?

sciwhiz12 (changes requested):
Mostly looks fine, just have some questions to clarify some terms in `SignalGetter#getControlInputSignal`.

data/net/minecraft/world/level/SignalGetter.mapping
SpaceWalkerRS: Something Mojang made up. I just took it from the method name and clarified it more in the rest of the comment. It can return specific types of redstone signals from specific blocks.

data/net/minecraft/world/level/SignalGetter.mapping
SpaceWalkerRS: That's correct. Yeah sounds good, I'll go add a link to that comment.

data/net/minecraft/world/level/SignalGetter.mapping
sciwhiz12: Eh, Mojang do what Mojang do (or did).

While I can't for certain conclude it, I'm getting the feeling (based on some quick looks on calls to the method) that this is related to the input signals that can be fed to comparators. I'm guessing their definition of an 'control signal' is a signal which can control the behavior of some redstone-emitting block, such as a comparator (diode) block, but that's probably something to investigate for another time.

sciwhiz12 (approved):

sciwhiz12:
Just a note: in the future, you'll want to make a separate branch for your PRs instead of PRing the version branches directly, to avoid the hassle of having to fix your version branch to match upstream (this repository) again.

data/net/minecraft/world/level/SignalGetter.mapping
SpaceWalkerRS: iirc the method is used by repeaters/comparators for their side input, and also the calibrated sculk sensor, to make it listen to specific "frequencies" - the question is always *is it about what the method is used for, or what it does?* and in this case, because the method is surrounded by other "generic" signal-getter methods, I opted to go for the latter.

SpaceWalkerRS:
thanks for the advice, I'll keep it in mind

PR #204, opened 2023-06-27
Update to Gradle 8

Matyrobbrt:
The foojay plugin is added because Enigma runs on Java 17


marchermans (dismissed):

sciwhiz12 (changes requested):
There seems to be changes when I run `./gradlew wrapper --gradle-version=8.1.1` on this PR.

Notably, the checksum of the Gradle wrapper JAR in this repository doesn't matches the checksum for Gradle 8.1.1, but it does match the wrapper checksum for 7.6.1. Please rerun the above task to update the wrapper JAR to the correct one for 8.1.1.

sciwhiz12 (approved):

PR #205, opened 2023-06-29
Add some blaze3d and com.mojang.math mappings

Matyrobbrt:


data/com/mojang/math/Axis.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS com/mojang/math/Axis
+	METHOD of (Lorg/joml/Vector3f;)Lcom/mojang/math/Axis;
+		ARG 0 axis
+	METHOD rotation (F)Lorg/joml/Quaternionf;
+		ARG 1 angle
```
sciwhiz12: Calling this `radians` would make it clear that it accepts an angle measurement in radians, as opposed to `rotationDegrees` which accepts an angle measurement in degrees.

```suggestion
		ARG 1 radians
```

sciwhiz12 (changes requested):

sciwhiz12 (approved):

PR #24, opened 2021-06-25
Parameter names and Javadocs for net.minecraft.network and children

diesieben07:

PR #207, opened 2023-06-29
Map classes in net.minecraft.util

Matyrobbrt:


data/net/minecraft/util/Crypt.mapping
```diff
@@ -34,8 +36,26 @@ CLASS net/minecraft/util/Crypt
...
 		ARG 2 key
+	METHOD stringToPemRsaPrivateKey (Ljava/lang/String;)Ljava/security/PrivateKey;
+		ARG 0 keyBase64
+	METHOD stringToRsaPublicKey (Ljava/lang/String;)Ljava/security/PublicKey;
+		ARG 0 keyBase64
+	CLASS SaltSignaturePair
+		METHOD <init> (Lnet/minecraft/network/FriendlyByteBuf;)V
+			ARG 1 buf
+		METHOD write (Lnet/minecraft/network/FriendlyByteBuf;Lnet/minecraft/util/Crypt$SaltSignaturePair;)V
+			ARG 0 buf
```
alcatrazEscapee: `buf` -> `buffer`, imo, as per [Standards](https://parchmentmc.org/docs/mappings) Mappings 4. "Avoid abbreviations or acronyms."

A cursory search through the repo indicates we have in multiple places, both `buffer` and `buf` used already for `FriendlyByteBuf` parameters.

SizableShrimp: Is `buf` not valid even if it's already part of the classname?

alcatrazEscapee: imo, no, that doesn't pass the smell test as to if we should use the abbreviation/class name vs. prefer the full word. Which is consistent with past sentiment on the matter, i.e. [here](https://github.com/ParchmentMC/Parchment/pull/24#issuecomment-868975548) where we standardized on `buffer` (vs. `buf` or `byteBuf`)

alcatrazEscapee (approved):

PR #221, opened 2023-07-14
Various mappings in net/minecraft/util

PiTheGuy:


data/net/minecraft/resources/HolderSetCodec.mapping
```diff
@@ -1,8 +1,18 @@
...
 	METHOD create (Lnet/minecraft/resources/ResourceKey;Lcom/mojang/serialization/Codec;Z)Lcom/mojang/serialization/Codec;
 		ARG 0 registryKey
 		ARG 1 holderCodec
 		ARG 2 disallowInline
+	METHOD decode (Lcom/mojang/serialization/DynamicOps;Ljava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 1 ops
+		ARG 2 value
+	METHOD encode (Lnet/minecraft/core/HolderSet;Lcom/mojang/serialization/DynamicOps;Ljava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 2 ops
+		ARG 3 value
```
alcatrazEscapee: Args should be `input`, `ops`, `prefix`, as per the parameters on [`Encoder`](https://github.com/Mojang/DataFixerUpper/blob/6d0d1e6281abd94328b135b53f3f3b592dc9f6cd/src/main/java/com/mojang/serialization/Encoder.java#L11)

data/net/minecraft/resources/HolderSetCodec.mapping
```diff
@@ -1,8 +1,18 @@
...
+		ARG 1 registryKey
+		ARG 2 elementCodec
+		ARG 3 disallowInline
 	METHOD create (Lnet/minecraft/resources/ResourceKey;Lcom/mojang/serialization/Codec;Z)Lcom/mojang/serialization/Codec;
 		ARG 0 registryKey
 		ARG 1 holderCodec
 		ARG 2 disallowInline
+	METHOD decode (Lcom/mojang/serialization/DynamicOps;Ljava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 1 ops
+		ARG 2 value
```
alcatrazEscapee: Should be `input`, as per the interface [`Decoder`](https://github.com/Mojang/DataFixerUpper/blob/6d0d1e6281abd94328b135b53f3f3b592dc9f6cd/src/main/java/com/mojang/serialization/Decoder.java#L14)

data/net/minecraft/util/CubicSpline.mapping
```diff
@@ -18,6 +18,9 @@ CLASS net/minecraft/util/CubicSpline
 		METHOD <init> (Lnet/minecraft/util/ToFloatFunction;Lnet/minecraft/util/ToFloatFunction;)V
 			ARG 1 coordinate
 			ARG 2 valueTransformer
+		METHOD addPoint (FF)Lnet/minecraft/util/CubicSpline$Builder;
+			ARG 1 location
+			ARG 2 calue
```
alcatrazEscapee:
```suggestion
			ARG 2 value
```

data/net/minecraft/util/Graph.mapping
```diff
@@ -0,0 +1,7 @@
+CLASS net/minecraft/util/Graph
+	METHOD depthFirstSearch (Ljava/util/Map;Ljava/util/Set;Ljava/util/Set;Ljava/util/function/Consumer;Ljava/lang/Object;)Z
+		ARG 0 graph
+		ARG 1 visitedVerticies
+		ARG 2 recursionStack
+		ARG 3 action
+		ARG 4 currentVertex
```
alcatrazEscapee:
```suggestion
	METHOD depthFirstSearch (Ljava/util/Map;Ljava/util/Set;Ljava/util/Set;Ljava/util/function/Consumer;Ljava/lang/Object;)Z
		ARG 0 graph
		ARG 1 nonCyclicalNodes
		ARG 2 pathSet
		ARG 3 onNonCyclicalNodeFound
		ARG 4 currentNode
```

This is a method used purely by the feature cycle sorter and it's used for a specific purpose - to identify cycles and (unfortunately) not return them. I have had for a while, a [mapped + commented](https://github.com/alcatrazEscapee/cyanide/blob/517557efb96c6b44282d269a91317ebd648debce/Common/src/main/java/com/alcatrazescapee/cyanide/codec/FeatureCycleDetector.java#L213) version of this method in Cyanide, and would suggest the names inherited from there.

It would also be nice to add some comments to this effect:
```java
/**
 * Detects if a cycle is present in the given graph, via a depth first search, and returns {@code true} if a cycle was found.
 * 
 * @param nonCyclicalNodes Nodes that are verified to have no cycles involving them.
 * @param pathSet The current collection of seen nodes. When invoked not recursively, this should be an empty set.
 * @param onNonCyclicalNodeFound Invoked on each node as we prove that no cycles can be reached starting from this node.
```

alcatrazEscapee (changes requested):

alcatrazEscapee (approved):

PR #233, opened 2023-07-29
Fix name collision in RegistrySetBuilder

Machine-Maker:
The parameter cannot be named `owner` because it collides with a visible field named `owner` inside `EmptyTagLookup`.


SizableShrimp:
Name collisions are not a concern. If you are recompiling the game, you should use the special release artifact which prefixes each parameter with "p" (e.g. used for Forge) If you are not recompiling the game, then it is not a concern and you should use the normal release as it's a purely visual/IDE problem (e.g. Fabric). Ask on the Discord for more help.

PR #234, opened 2023-07-30
finish params for world.level.storage.loot.functions

Machine-Maker:
I think some of the new deserialize methods are identical to the super class' method, so if you don't want them duplicated I can take them out.


alcatrazEscapee:
> I think some of the new deserialize methods are identical to the super class' method

IF you run `sanitizeData`, it should remove duplicate (by class hierarchy) comments/parameters, I believe, because they are populated on export anyway.

Lanse505 (approved):

PR #243, opened 2023-08-03
finish mapping world/level/biome

Machine-Maker:


data/net/minecraft/world/level/biome/BiomeGenerationSettings.mapping
```diff
@@ -36,3 +36,7 @@ CLASS net/minecraft/world/level/biome/BiomeGenerationSettings
 			ARG 2 feature
 		METHOD addFeatureStepsUpTo (I)V
 			ARG 1 step
+		METHOD lambda$addCarver$0 (Lnet/minecraft/world/level/levelgen/GenerationStep$Carving;)Ljava/util/List;
+			ARG 0 ignored
```
alcatrazEscapee:
```suggestion
			ARG 0 key
```

It's a `map.computeIfAbsent(key -> new ArrayList())`, which reads better than "ignored".

data/net/minecraft/world/level/biome/BiomeGenerationSettings.mapping
```diff
@@ -36,3 +36,7 @@ CLASS net/minecraft/world/level/biome/BiomeGenerationSettings
...
+		METHOD lambda$build$1 (Ljava/util/Map$Entry;)Lnet/minecraft/core/HolderSet;
+			ARG 0 carverEntry
```
alcatrazEscapee:
```suggestion
			ARG 0 carvers
```

data/net/minecraft/world/level/biome/FeatureSorter.mapping
```diff
@@ -1,4 +1,16 @@
 CLASS net/minecraft/world/level/biome/FeatureSorter
+	METHOD buildFeaturesPerStep (Ljava/util/List;Ljava/util/function/Function;Z)Ljava/util/List;
+		ARG 0 featureSetSources
+		ARG 1 toFeatueSetFunction
+		ARG 2 notRecursiveFlag
+	METHOD lambda$buildFeaturesPerStep$0 (Lorg/apache/commons/lang3/mutable/MutableInt;Ljava/lang/Object;)I
+		ARG 1 ignored
+	METHOD lambda$buildFeaturesPerStep$1 (Ljava/util/Comparator;Lnet/minecraft/world/level/biome/FeatureSorter$1FeatureData;)Ljava/util/Set;
+		ARG 1 ignored
```
alcatrazEscapee:
```suggestion
		ARG 1 key
	METHOD lambda$buildFeaturesPerStep$1 (Ljava/util/Comparator;Lnet/minecraft/world/level/biome/FeatureSorter$1FeatureData;)Ljava/util/Set;
		ARG 1 key
```

Same as above, I think this reads better given it's `.computeIfAbsent(key -> new thing)`

alcatrazEscapee (changes requested):

data/net/minecraft/world/level/biome/BiomeGenerationSettings.mapping
Machine-Maker: I was kinda thinking that deliberately ignored params would be better as `ignored` to signify that they were deliberately ignored.

alcatrazEscapee: I'd agree, if the parameter was otherwise unknown as to *why* it's ignored. E.g.

```java
// 'ignored' tells the caller they can pass whatever they want in here, which is useful information
void methodAModderMayWantToCall(int ignored) { ... }
```

But in this case, we're looking at the parameter in a lambda within a MC method. Which makes more sense to name it based on what is the most readable, in this case, `computeIfAbsent` takes a `key` and returns a new mapping.

```
map.computeIfAbsent(key -> new ArrayList<>()); // this reads better
map.computeIfAbsent(ignored -> new ArrayList<>()); // what does being `ignored` tell me, that is immediately obvious in this case.
```

Anecdotally, I've seen this exact pattern (with `_k`, `k`, `key` or similar), way more over a MC codebase + libraries than `ignored`.

Machine-Maker: yep, ok sure.

alcatrazEscapee (approved):

PR #244, opened 2023-08-03
Consistency for `RecordCodecBuilder$Instance`

alcatrazEscapee:
`codec` is a bad name, it's not the codec, other things are bad and/or too long and `instance` is the functional-wanker-y-correct name.


Lanse505 (dismissed):
Approved but needs Rebase

sciwhiz12 (approved):

PR #246, opened 2023-08-03
finish mapping world/level/block/entity

Machine-Maker:
If someone's got a better name for that bool than `dontPlaceOnSizeMismatch`, please say so. That seems to be it's functionality if the size of the structure doesn't match the template, and that is set to true, it won't place it in the world.

### DIFF
--- a/data/net/minecraft/gametest/framework/StructureUtils.mapping
+++ b/data/net/minecraft/gametest/framework/StructureUtils.mapping
@@ -23,6 +23,7 @@ CLASS net/minecraft/gametest/framework/StructureUtils
 		ARG 1 pos
 		ARG 2 rotation
 		ARG 3 serverLevel
+		ARG 4 dontPlaceOnSizeMismatch
 	METHOD doesStructureContain (Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/BlockPos;Lnet/minecraft/server/level/ServerLevel;)Z
 		ARG 0 structureBlockPos
 		ARG 1 posToTest
@@ -74,5 +75,6 @@ CLASS net/minecraft/gametest/framework/StructureUtils
 		ARG 1 pos
 		ARG 2 rotation
 		ARG 4 serverLevel
+		ARG 5 dontPlaceOnSizeMismatch
 	METHOD tryLoadStructure (Ljava/nio/file/Path;)Lnet/minecraft/nbt/CompoundTag;
 		ARG 0 pathToStructure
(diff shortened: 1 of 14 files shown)


data/net/minecraft/gametest/framework/StructureUtils.mapping
```diff
@@ -74,5 +75,6 @@ CLASS net/minecraft/gametest/framework/StructureUtils
 		ARG 1 pos
 		ARG 2 rotation
 		ARG 4 serverLevel
+		ARG 5 dontPlaceOnSizeMismatch
```
Matyrobbrt, line 78: Could be `ignoreSizeMismatch` instead?

data/net/minecraft/world/level/block/entity/TheEndGatewayBlockEntity.mapping
```diff
@@ -34,6 +34,10 @@ CLASS net/minecraft/world/level/block/entity/TheEndGatewayBlockEntity
 	METHOD isChunkEmpty (Lnet/minecraft/server/level/ServerLevel;Lnet/minecraft/world/phys/Vec3;)Z
 		ARG 0 level
 		ARG 1 pos
+	METHOD lambda$findOrCreateValidTeleportPos$0 (Lnet/minecraft/core/Registry;)Ljava/util/Optional;
+		ARG 0 configuredFeatureRegistry
+	METHOD lambda$findOrCreateValidTeleportPos$1 (Lnet/minecraft/server/level/ServerLevel;Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/Holder$Reference;)V
+		ARG 2 featureHolder
```
Matyrobbrt:
```suggestion
		ARG 2 feature
```

Matyrobbrt (dismissed):

data/net/minecraft/gametest/framework/StructureUtils.mapping
Machine-Maker: It's the opposite of that. If it's true, it doesn't ignore it. If it's false, it does ignore the mismatch.

SizableShrimp: failOnSizeMismatch?

Machine-Maker: I thought about that too, but it's not really failing. Like some logic does run anyways. idk.

alcatrazEscapee: In this case, it might be useful to add a `@param` comment for this parameter explaining exactly what this means, as it's clearly not obvious just via name (and I don't think that it will be via the choice of the best possible flag name).

Explanatory comment > bikshed parameter name.

alcatrazEscapee (commented):
A explanatory `@param` comment for the `dontPlaceOnSizeMismatch` parameters would be much appreciated, otherwise I have no other comments for the rest of the PR.

data/net/minecraft/gametest/framework/StructureUtils.mapping
Machine-Maker: Ok, added a comment.

alcatrazEscapee (approved):

PR #251, opened 2023-08-05
finish mapping world/level/material,pathfinder,portal,saveddata,valid…

Machine-Maker:
…ation

### DIFF
--- a/data/net/minecraft/world/level/pathfinder/SwimNodeEvaluator.mapping
+++ b/data/net/minecraft/world/level/pathfinder/SwimNodeEvaluator.mapping
@@ -18,7 +18,7 @@ CLASS net/minecraft/world/level/pathfinder/SwimNodeEvaluator
 	METHOD isDiagonalNodeValid (Lnet/minecraft/world/level/pathfinder/Node;Lnet/minecraft/world/level/pathfinder/Node;Lnet/minecraft/world/level/pathfinder/Node;)Z
 		ARG 1 root
 		ARG 2 horizontal
-		ARG 3 clockWise
+		ARG 3 clockwise
 	METHOD isNodeValid (Lnet/minecraft/world/level/pathfinder/Node;)Z
 		ARG 1 node
 	METHOD lambda$getCachedBlockType$0 (IIIJ)Lnet/minecraft/world/level/pathfinder/BlockPathTypes;
--- a/data/net/minecraft/world/level/pathfinder/SwimNodeEvaluator.mapping
+++ b/data/net/minecraft/world/level/pathfinder/SwimNodeEvaluator.mapping
@@ -15,5 +15,11 @@ CLASS net/minecraft/world/level/pathfinder/SwimNodeEvaluator
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
+	METHOD isDiagonalNodeValid (Lnet/minecraft/world/level/pathfinder/Node;Lnet/minecraft/world/level/pathfinder/Node;Lnet/minecraft/world/level/pathfinder/Node;)Z
+		ARG 1 root
+		ARG 2 horizontal
+		ARG 3 clockWise
 	METHOD isNodeValid (Lnet/minecraft/world/level/pathfinder/Node;)Z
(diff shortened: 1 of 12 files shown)


data/net/minecraft/world/level/pathfinder/SwimNodeEvaluator.mapping
```diff
@@ -15,5 +15,11 @@ CLASS net/minecraft/world/level/pathfinder/SwimNodeEvaluator
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
+	METHOD isDiagonalNodeValid (Lnet/minecraft/world/level/pathfinder/Node;Lnet/minecraft/world/level/pathfinder/Node;Lnet/minecraft/world/level/pathfinder/Node;)Z
+		ARG 1 root
+		ARG 2 horizontal
+		ARG 3 clockWise
```
alcatrazEscapee:
```suggestion
		ARG 3 clockwise
```

data/net/minecraft/world/level/validation/DirectoryValidator.mapping
```diff
@@ -3,16 +3,17 @@ CLASS net/minecraft/world/level/validation/DirectoryValidator
...
+			ARG 2 attrs
 		METHOD validateSymlink (Ljava/nio/file/Path;Ljava/nio/file/attribute/BasicFileAttributes;)V
 			ARG 1 path
-			ARG 2 attributes
+			ARG 2 attrs
 		METHOD visitFile (Ljava/nio/file/Path;Ljava/nio/file/attribute/BasicFileAttributes;)Ljava/nio/file/FileVisitResult;
-			ARG 1 path
-			ARG 2 attributes
+			ARG 1 file
+			ARG 2 attrs
```
alcatrazEscapee: @ other reviewers: these parameters are copied from `java.nio.file.SimpleFileVisitor`. Personally, I think the existing ones are better (i.e. `attributes` over `attrs`) and more consistent with our standard. But it is different from the super-class parameters.

My vote is to keep `path`, `attributes`, etc.

alcatrazEscapee (changes requested):

data/net/minecraft/world/level/validation/DirectoryValidator.mapping
Machine-Maker: While keeping `attributes` over `attrs` is fine, I think `file` and `dir` are better and convey more information. A Path could be either, but the Path passed as an argument will be a file or directory afaik. I then think its better to keep the specific `file` and `dir` names.

PiTheGuy: I agree with MachineMaker on this. I prefer `attributes` over `attrs`, but think `file` and `dir` are better.

Lanse505: 2 Cents:
attributes -> attr  | For readability
file & dir               | For readability

Question is if we should plan a sit down to make a pool of standardized names that we then apply over the entire project like we did before with tileEntity -> blockEntity and such?

alcatrazEscapee: I'm ambivalent to `file` / `dir`. My main comment is, in keeping with Parchment standards, to avoid abbreviations or shortenings (like `attrs` or `dir`) instead of the actual word (`attributes`, `directory`).

It's not a matter of having a standardized naming pass; it's a matter of sticking to the existing standards.

> 4. Avoid abbreviations or acronyms.

Lanse505: The thing is also while yes "dir" is an abbreviation, it's such a established term in programming as a whole that I'm not sure I'd consider it a "pure abbreviation" rather than an established contextual word.

I can honestly go either way with dir vs directory so /shrug

Machine-Maker: So currently it's using `attributes` and `path` and `dir`.

alcatrazEscapee (approved):

PR #256, opened 2023-08-18
Fix incorrect formatting preventing Enigma from correctly loading mapping files

PiTheGuy:


Lanse505 (approved):
This is funny since the comments are generated by either Scribe or Enigma.
We might want to look into if its Scribe or Enigma causing this for the future but for now this looks fine.

I'll /cc @SizableShrimp @sciwhiz12 in-case they're intressted, but for now I'll approve and merge.

PiTheGuy:
I think this was from me manually editing the mapping files to fix a Javadoc error in my previous pull request (#254). It might still be a good idea to double check but I'm not sure it's too big of a concern. I've fixed my formatting settings in IntelliJ so this specific issue should happen again.

PR #257, opened 2023-08-18
Fix incorrect blitNineSliced interpretation

JackFred2:


Lanse505 (approved):
That's my bad sorry, though do note all of this code got removed in 1.20.2 soooooo sorry for the news :P

PR #263, opened 2023-10-31
Mappings in net/minecraft/world/level

PiTheGuy:


sciwhiz12 (approved):
I currently don't have the means to do my usual cross-checking of these mapping names in the actual code, so I'm opting to trust your judgement and instead look for typos for now.

Having said that, I found no typos. 😄

PR #283, opened 2024-06-13
Added Javadoc and changed parameter names in relation to `ChunkTrackingView` (fixed inaccurate javadoc)

FamroFexl:
Javadoc was provided for several ChunkTrackingView methods.
Several parameter names were changed or flipped to better reflect their purpose.

Changing the `searchAllChunks` parameter names in the `ChunkTrackingView` class triggered hundreds of edits in commit b0041f886524887058babf09bd15a09928e2ef99 from enigma for some reason.

### DIFF
--- a/data/com/mojang/blaze3d/pipeline/MainTarget.mapping
+++ b/data/com/mojang/blaze3d/pipeline/MainTarget.mapping
@@ -12,9 +12,6 @@ CLASS com/mojang/blaze3d/pipeline/MainTarget
 	METHOD createFrameBuffer (II)V
 		ARG 1 width
 		ARG 2 height
-	CLASS AttachmentState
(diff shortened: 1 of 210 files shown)


PiTheGuy:
A recent enigma update changed the order in which mappings are saved. Run the `santinizeData` gradle task and it should fix it.

FamroFexl:
`sanitizeData` task run in commit 7bf753ec733f55930069e1e5f12459d7ed968f52

sciwhiz12 (approved):

sciwhiz12:
Thanks! We'll look into the changed ordering written out by Enigma and match ours in the future. (It seems simple at least -- the order of inner classes is now sorted rather than being in encounter order from the bytecode, I believe.)

PR #288, opened 2024-06-26
Mapping for `advancements` and `client` packages

PiTheGuy:


data/net/minecraft/client/renderer/LevelRenderer.mapping
```diff
@@ -34,8 +34,8 @@ CLASS net/minecraft/client/renderer/LevelRenderer
 		ARG 10 blue
 		ARG 11 alpha
 	METHOD addFrustumQuad (Lcom/mojang/blaze3d/vertex/VertexConsumer;Lorg/joml/Matrix4f;IIIIIII)V
-		ARG 1 consumer
-		ARG 2 matrix
+		ARG 1 buffer
```
sciwhiz12: Why the change from `consumer` to `buffer` in these methods, since the type is `VertexConsumer`?

data/net/minecraft/client/renderer/entity/LightningBoltRenderer.mapping
```diff
@@ -3,8 +3,8 @@ CLASS net/minecraft/client/renderer/entity/LightningBoltRenderer
 		COMMENT Returns the location of an entity's texture.
 		ARG 1 entity
 	METHOD quad (Lorg/joml/Matrix4f;Lcom/mojang/blaze3d/vertex/VertexConsumer;FFIFFFFFFFZZZZ)V
-		ARG 0 matrix
-		ARG 1 consumer
+		ARG 0 pose
+		ARG 1 buffer
```
sciwhiz12: RE: Why the change from `consumer` to `buffer` for type `VertexConsumer`?

data/net/minecraft/client/renderer/LevelRenderer.mapping
PiTheGuy: `buffer` is more correct. The `VertexConsumer` is always sourced from `MultiBufferSource.getBuffer()` and represents a buffer.

data/net/minecraft/client/renderer/entity/LightningBoltRenderer.mapping
PiTheGuy: `buffer` is more correct. The `VertexConsumer` is always sourced from `MultiBufferSource.getBuffer()` and represents a buffer.

data/net/minecraft/client/renderer/LevelRenderer.mapping
sciwhiz12: From the perspective of that method, it has no distinction whether the caller gives it a buffer or not. I think it should still be `consumer`, since it is possible something else (a mod, perhaps) can call that method with a `VertexConsumer` that isn't necessarily a buffer from `MultiBufferSource`.

PiTheGuy:
I do not feel it is worth the time and effort to resolve merge conflicts on this PR when all of its mappings have since been added in other PRs.

PR #296, opened 2024-08-05
fix some incorrect mappings

BasiqueEvangelist:
- ServerPlayer#sendSystemMessage's `bypassHiddenChat` is renamed to `overlay`, since it also controls whether the message goes to the action bar
- closeContainer and doCloseContainer's javadocs are changed to better reflect what they do
- Container's invalid stillValid javadoc is removed and some other javadocs are changed to fit the methods more

### DIFF
--- a/data/net/minecraft/server/level/ServerPlayer.mapping
+++ b/data/net/minecraft/server/level/ServerPlayer.mapping
@@ -11,7 +11,7 @@ CLASS net/minecraft/server/level/ServerPlayer
 		ARG 3 gameProfile
 		ARG 4 clientInformation
 	METHOD acceptsSystemMessages (Z)Z
-		ARG 1 bypassHiddenChat
+		ARG 1 overlay
 	METHOD addAdditionalSaveData (Lnet/minecraft/nbt/CompoundTag;)V
 		ARG 1 compound
 	METHOD attack (Lnet/minecraft/world/entity/Entity;)V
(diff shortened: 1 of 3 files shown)


data/net/minecraft/server/level/ServerPlayer.mapping
```diff
@@ -74,7 +74,7 @@ CLASS net/minecraft/server/level/ServerPlayer
 		ARG 5 movementZ
 		ARG 7 onGround
 	METHOD doCloseContainer ()V
-		COMMENT Closes the container the player currently has open.
+		COMMENT Invoked when the container the player has opened is to be closed, either by a call to {@link #closeContainer} or a packet.
```
Matyrobbrt:
```suggestion
		COMMENT Invoked when the active container of the player is to be closed, either by a call to {@link #closeContainer()} or a packet.
```
I feel like this wording is better

data/net/minecraft/server/level/ServerPlayer.mapping
```diff
@@ -51,7 +51,7 @@ CLASS net/minecraft/server/level/ServerPlayer
 		ARG 3 dy
 		ARG 5 dz
 	METHOD closeContainer ()V
-		COMMENT Sets the current crafting inventory back to the 2x2 square.
+		COMMENT Closes the container the player currently has open.
```
Matyrobbrt, line 54: Same as below

data/net/minecraft/world/Container.mapping
```diff
@@ -38,15 +38,14 @@ CLASS net/minecraft/world/Container
 		COMMENT Removes a stack from the given slot and returns it.
 		ARG 1 slot
 	METHOD setChanged ()V
-		COMMENT For block entities, ensures the chunk containing the block entity is saved to disk later - the game won't think it hasn't changed and skip it.
+		COMMENT Called whenever this container is changed.
```
Matyrobbrt:
```suggestion
		COMMENT Called when the contents of this container have changed.
```

Matyrobbrt (changes requested):

Lanse505 (approved):
LGTM

PR #308, opened 2024-11-03
Rename RandomSource.triangle arguments

Sleepy-Horse:
The code of the function is
```
default double triangle(double min, double max) {
        return min + max * (this.nextDouble() - this.nextDouble());
    }
```
It can return `min - max` or `min + max`. I believe the arguments should be named `center` and `maxDeviation`.


sciwhiz12 (approved):
Ah, because `nextDouble()` returns a value from 0.0 to 1.0 in each call, the value of the multiplier to `max` can range from 1.0 (`1.0 - 0.0`) to -1.0 (`0.0 - 1.0`), or thereabouts when accounting for 1.0 being the exclusive range (rather than inclusive). Looks fine to me, then.

PR #309, opened 2024-11-09
Add GitHub Actions to replace TeamCity as CI

sciwhiz12:
As many of us are (painfully) aware, a few months ago, the LDTTeam TeamCity was shut down due to various reasons such as operating costs. That TeamCity instance ran all of the CI for ParchmentMC, so work on Parchment mappings ground to a standstill until we replaced the now-dead TeamCity with another CI solution.

Of course, the natural replacement (and the one that LDTTeam ran with, even) was [GitHub Actions](https://github.com/features/actions), an on-platform integrated CI solution running off YAML files stored in the repository, which GitHub grants free access to public repositories. The majority of projects in the Minecraft modding space (or at least, those I know of) use GitHub Actions for their build and release workflows, due to its simplicity, ease of use, and availability (since it's available for any repository on GitHub).

What remained was writing up the GitHub Action files to replace the TeamCity build configurations. As toolchain lead, it is my responsibility to do that, partly because without access to the old TeamCity instance, I'm nearly the only one with the (semi-)working memory on what the build configurations even did in the first place. That compounded with my own personal/academic life, and general pain in figuring out what best to start with for writing the GitHub Actions.

But after all that, we (finally!) now have the replacement GitHub Actions workflows for the TeamCity build configurations. I've looked at them twice now, and they _look_ fine to me, but the true test will of course be in production. (Cue "testing in production" memes here.) I've also worked with @marchermans to get the credentials to the LDTTeam JFrog (which powers our Maven repository) fixed up and ready.

---

What's left after this PR? Fixing bugs with the workflows if any (though hopefully not), reviewing all of the backlogged PRs starting from oldest to newest, and then working on making GitHub Action workflows for *all* the ParchmentMC repositories which relied on TeamCity previously.

Particularly, Blackstone is next on the list to be GitHub Action'd, since that is required for validation and sanitation for the new MC versions. However, I'll probably just run and publish Blackstone manually for the few versions that don't have one yet since TeamCity was shut down, and work on automating that within the next few months.

### DIFF
--- a/.github/workflows/bleeding.yml
+++ b/.github/workflows/bleeding.yml
@@ -1,0 +1,78 @@
+on: 
+  workflow_dispatch:
+  push:
+
+name: Publish bleeding export
+
+jobs:
+  publish:
+    name: Publish export
+    runs-on: ubuntu-latest
+    outputs:
+      CI_VERSION: ${{ steps.info.outputs.CI_VERSION }}
+      CI_GROUP: ${{ steps.info.outputs.CI_GROUP }}
+      CI_ARTIFACT: ${{ steps.info.outputs.CI_ARTIFACT }}
+      CI_GAME_VERSION: ${{ steps.info.outputs.CI_GAME_VERSION }}
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 17
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
+
+      - name: Obtain GAV info
+        id: info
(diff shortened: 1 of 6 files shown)


marchermans:
Very nice work!

sciwhiz12:
Ah, I just realized a current design flaw with the setup of the workflows. They're designed to work against a _single_ branch—the default branch—but won't work with our multiple branches for multiple versions. The variables would need to be made multi-branch aware (from a simple boolean to a JSON list of branch names as strings, perhaps), and the nightly and release workflows would need to be refactored a bit to run against each branch rather than just the default branch.

But we'll consider that as future work for now, because refactoring those takes a bit more mind power than I can muster at this hour, and this should be good enough to at least move forward with `1.21.x`.

PR #318, opened 2025-01-17
Fix parameter shadow in `DownloadCacheCleaner`

Matyrobbrt:
During the 1.20.4 port both the in path parameter and the parameter of the anonymous class method in `vacuumCacheDir` were named `path` therefore leading to the outer variable being shadowed by the inner one.


sciwhiz12 (approved):
Making a validator for _this_ kind of error is going to be painful, and possibly requires some rethinking of the design of our toolchain.

For now, hotfix!

PR #334, opened 2025-06-17
Update to 1.21.6

PiTheGuy:


data/com/mojang/blaze3d/opengl/BufferStorage.mapping
```diff
@@ -0,0 +1,26 @@
...
+		ARG 1 directStateAccess
+		ARG 2 buffer
+		ARG 3 offset
+		ARG 4 size
+		ARG 5 usage
+	CLASS Immutable
+		METHOD tryMapBufferPersistent (Lcom/mojang/blaze3d/opengl/DirectStateAccess;III)Ljava/nio/ByteBuffer;
+			ARG 1 directStateAccess
+			ARG 2 usage
+			ARG 3 buffer
```
Lanse505: This really should be ``handle`` looking at various calls using that, falling back into the ``GlBuffer`` field of ``handle``

Lanse505:
@RogueLogix 
This PR essentially adds the same things as your PR with some name differences for the Blaze3D stuff. If you want to comment on his choice of names and start a discussion feel free to :)

I'm adding a few comments where I find it suitable for now.

data/net/minecraft/client/KeyboardHandler.mapping
```diff
@@ -24,12 +24,22 @@ CLASS net/minecraft/client/KeyboardHandler
...
+	METHOD debugFeedbackFormatted (Ljava/lang/String;[Ljava/lang/Object;)V
+		ARG 1 message
+		ARG 2 args
+	METHOD debugFeedbackTranslated (Ljava/lang/String;)V
+		ARG 1 message
 	METHOD debugFeedbackTranslated (Ljava/lang/String;[Ljava/lang/Object;)V
 		ARG 1 message
 		ARG 2 args
+	METHOD debugWarningComponent (Lnet/minecraft/network/chat/Component;)V
+		ARG 1 mesage
```
Lanse505: ``mesage`` -> ``message``

data/net/minecraft/client/Minecraft.mapping
```diff
@@ -11,6 +11,8 @@ CLASS net/minecraft/client/Minecraft
 		ARG 3 registryAccess
 	METHOD addInitialScreens (Ljava/util/List;)V
 		ARG 1 output
+	METHOD addInitialScreens (Ljava/util/List;)Z
+		ARG 1 outut
```
Lanse505: ``outut`` -> ``output``

Lanse505 (changes requested):
Looks mostly good to me, there are 3 small things I found which are quick fixes.

Lanse505 (dismissed):

Xyndra:
Hi, I would like to develop with parchment for the latest version, I don't know the ETA on how long it will take to get merged. Could you maybe tell me how to use the PR temporarily

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
```diff
@@ -4,6 +4,15 @@ CLASS com/mojang/blaze3d/systems/CommandEncoder
 		ARG 2 color
 		ARG 3 depthTexture
 		ARG 4 clearDepth
+	METHOD clearColorAndDepthTextures (Lcom/mojang/blaze3d/textures/GpuTexture;ILcom/mojang/blaze3d/textures/GpuTexture;DIIII)V
+		ARG 1 colorTextureAtlas
```
RogueLogix: this should be just "colorTexture", this is generic and not atlas specific. same for the depth one below

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
```diff
@@ -4,6 +4,15 @@ CLASS com/mojang/blaze3d/systems/CommandEncoder
...
+		ARG 2 color
+		ARG 3 depthTextureAtlas
+		ARG 4 clearDepth
+		ARG 6 x
```
RogueLogix: these should specify that they are scissor x/y/width/height

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
```diff
@@ -4,6 +4,15 @@ CLASS com/mojang/blaze3d/systems/CommandEncoder
 		ARG 2 color
 		ARG 3 depthTexture
 		ARG 4 clearDepth
+	METHOD clearColorAndDepthTextures (Lcom/mojang/blaze3d/textures/GpuTexture;ILcom/mojang/blaze3d/textures/GpuTexture;DIIII)V
+		ARG 1 colorTextureAtlas
+		ARG 2 color
```
RogueLogix: clearColor

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
```diff
@@ -44,8 +53,38 @@ CLASS com/mojang/blaze3d/systems/CommandEncoder
 		ARG 2 color
 		ARG 3 depthTexture
 		ARG 4 clearDepth
+	METHOD createRenderPass (Ljava/util/function/Supplier;Lcom/mojang/blaze3d/textures/GpuTexture;Ljava/util/OptionalInt;)Lcom/mojang/blaze3d/systems/RenderPass;
+		ARG 1 debugGroup
```
RogueLogix: "label" is the name Mojang uses for this afaict.

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
```diff
@@ -59,6 +98,9 @@ CLASS com/mojang/blaze3d/systems/CommandEncoder
 		ARG 1 buffer
 		ARG 2 source
 		ARG 3 offset
+	METHOD writeToBuffer (Lcom/mojang/blaze3d/buffers/GpuBufferSlice;Ljava/nio/ByteBuffer;)V
+		ARG 1 slice
+		ARG 2 buffer
```
RogueLogix: "data" would be consistent with the LWJGL param names

data/com/mojang/blaze3d/systems/GpuDevice.mapping
```diff
@@ -1,4 +1,12 @@
 CLASS com/mojang/blaze3d/systems/GpuDevice
+	METHOD createBuffer (Ljava/util/function/Supplier;II)Lcom/mojang/blaze3d/buffers/GpuBuffer;
+		ARG 1 label
+		ARG 2 usage
+		ARG 3 size
+	METHOD createBuffer (Ljava/util/function/Supplier;ILjava/nio/ByteBuffer;)Lcom/mojang/blaze3d/buffers/GpuBuffer;
+		ARG 1 label
+		ARG 2 usage
+		ARG 3 byteBuffer
```
RogueLogix: "data" again

data/com/mojang/blaze3d/systems/RenderPass.mapping
```diff
@@ -2,30 +2,61 @@ CLASS com/mojang/blaze3d/systems/RenderPass
...
+		ARG 1 name
+		ARG 2 texture
 	METHOD draw (II)V
 		ARG 1 firstIndex
 		ARG 2 indexCount
 	METHOD drawIndexed (II)V
 		ARG 1 firstIndex
 		ARG 2 indexCount
+	METHOD drawIndexed (III)V
+		ARG 1 firstIndex
```
RogueLogix: for consistency, all of the draw command names should be update to match the VK spec's names
VK and GL are both inconsistent within themselves, but VK is the more consistent one.

data/com/mojang/blaze3d/systems/RenderPass.mapping
```diff
@@ -2,30 +2,61 @@ CLASS com/mojang/blaze3d/systems/RenderPass
...
+		ARG 4 primCount
 	METHOD drawMultipleIndexed (Ljava/util/Collection;)V
 		ARG 1 draws
 	METHOD drawMultipleIndexed (Ljava/util/Collection;Lcom/mojang/blaze3d/buffers/GpuBuffer;Lcom/mojang/blaze3d/vertex/VertexFormat$IndexType;)V
 		ARG 1 draws
 		ARG 2 buffer
 		ARG 3 indexType
+	METHOD drawMultipleIndexed (Ljava/util/Collection;Lcom/mojang/blaze3d/buffers/GpuBuffer;Lcom/mojang/blaze3d/vertex/VertexFormat$IndexType;Ljava/util/Collection;)V
+		ARG 1 draws
+		ARG 2 buffer
```
RogueLogix: indexBuffer

data/com/mojang/blaze3d/systems/RenderPass.mapping
```diff
@@ -2,30 +2,61 @@ CLASS com/mojang/blaze3d/systems/RenderPass
...
+		ARG 3 indexType
+		ARG 4 uniforms
```
RogueLogix: dynamic uniform names

data/com/mojang/blaze3d/systems/RenderPass.mapping
```diff
@@ -2,30 +2,61 @@ CLASS com/mojang/blaze3d/systems/RenderPass
...
+	METHOD drawMultipleIndexed (Ljava/util/Collection;Lcom/mojang/blaze3d/buffers/GpuBuffer;Lcom/mojang/blaze3d/vertex/VertexFormat$IndexType;Ljava/util/Collection;Ljava/lang/Object;)V
+		ARG 1 draws
+		ARG 2 buffer
+		ARG 3 indexType
+		ARG 4 uniforms
+		ARG 5 data
```
RogueLogix: "userData" is the consistent with underlying APIs name

data/com/mojang/blaze3d/systems/RenderPass.mapping
```diff
@@ -2,30 +2,61 @@ CLASS com/mojang/blaze3d/systems/RenderPass
...
 		ARG 1 scissorState
+	METHOD pushDebugGroup (Ljava/util/function/Supplier;)V
+		ARG 1 name
 	METHOD setIndexBuffer (Lcom/mojang/blaze3d/buffers/GpuBuffer;Lcom/mojang/blaze3d/vertex/VertexFormat$IndexType;)V
 		ARG 1 indexBuffer
 		ARG 2 indexType
 	METHOD setPipeline (Lcom/mojang/blaze3d/pipeline/RenderPipeline;)V
 		ARG 1 pipeline
+	METHOD setUniform (Ljava/lang/String;Lcom/mojang/blaze3d/buffers/GpuBuffer;)V
+		ARG 1 uniform
```
RogueLogix: "name" to be consistent with existing setUniform calls

data/com/mojang/blaze3d/systems/RenderPass.mapping
```diff
@@ -47,6 +78,9 @@ CLASS com/mojang/blaze3d/systems/RenderPass
 			ARG 5 firstIndex
 			ARG 6 indexCount
 	CLASS UniformUploader
+		METHOD upload (Ljava/lang/String;Lcom/mojang/blaze3d/buffers/GpuBufferSlice;)V
+			ARG 1 uniform
```
RogueLogix: again, "name"

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
```diff
@@ -72,6 +114,17 @@ CLASS com/mojang/blaze3d/systems/CommandEncoder
 		ARG 7 height
 		ARG 8 sourceX
 		ARG 9 sourceY
+	METHOD writeToTexture (Lcom/mojang/blaze3d/textures/GpuTexture;Lcom/mojang/blaze3d/platform/NativeImage;IIIIIIII)V
+		ARG 1 texture
+		ARG 2 image
+		ARG 3 mipLevel
+		ARG 4 depthOrLayer
+		ARG 5 x
```
RogueLogix, line 79: while implied, this should probably specify that its dstX/Y

data/com/mojang/blaze3d/buffers/Std140SizeCalculator.mapping
RogueLogix, line 1: I believe this is incomplete

data/com/mojang/blaze3d/opengl/BufferStorage.mapping
```diff
@@ -0,0 +1,26 @@
+CLASS com/mojang/blaze3d/opengl/BufferStorage
+	METHOD create (Lorg/lwjgl/opengl/GLCapabilities;Ljava/util/Set;)Lcom/mojang/blaze3d/opengl/BufferStorage;
+		ARG 0 capabilities
+		ARG 1 enabledExtensions
+	METHOD createBuffer (Lcom/mojang/blaze3d/opengl/DirectStateAccess;Ljava/util/function/Supplier;II)Lcom/mojang/blaze3d/opengl/GlBuffer;
+		ARG 1 directStateAccess
```
RogueLogix, line 6: "dsa" is the name used elsewhere by Mojang, should this be consistent?

data/com/mojang/blaze3d/opengl/BufferStorage.mapping
```diff
@@ -0,0 +1,26 @@
...
+		ARG 2 label
+		ARG 3 usage
+		ARG 4 size
+	METHOD createBuffer (Lcom/mojang/blaze3d/opengl/DirectStateAccess;Ljava/util/function/Supplier;ILjava/nio/ByteBuffer;)Lcom/mojang/blaze3d/opengl/GlBuffer;
+		ARG 1 directStateAccess
+		ARG 2 label
+		ARG 3 usage
+		ARG 4 byteBuffer
```
RogueLogix: data

data/com/mojang/blaze3d/vertex/VertexConsumer.mapping
```diff
@@ -30,6 +30,11 @@ CLASS com/mojang/blaze3d/vertex/VertexConsumer
 		ARG 4 z
 	METHOD addVertex (Lorg/joml/Vector3f;)Lcom/mojang/blaze3d/vertex/VertexConsumer;
 		ARG 1 pos
+	METHOD addVertexWith2DPose (Lorg/joml/Matrix3x2f;FFF)Lcom/mojang/blaze3d/vertex/VertexConsumer;
+		ARG 1 vertex
```
RogueLogix: this feels wrong to me, should this not be "pose"?

RogueLogix (changes requested):
There's likely more things I've missed, but I don't have enigma or IDEA open at the moment to check in code.

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
PiTheGuy: Where are you seeing this? I looked through all the calls to it and didn't find any reference to it.

The current name comes from the following line: 
```java
this.device.debugLabels().pushDebugGroup(debugGroup);
```

data/com/mojang/blaze3d/buffers/Std140SizeCalculator.mapping
PiTheGuy: Engima shows 100% completion so unless you can tell me exactly what's missing, I'm pretty sure it's complete.

data/com/mojang/blaze3d/opengl/BufferStorage.mapping
PiTheGuy: Mojang isn't even consistent. Just to give an example the `DirectStateAccess` field in `GlDevice` is named `directStateAccess` not `dsa`

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
PiTheGuy: This sort of thing happens all the time in the `GuiGraphics` class and it's always specified as just `x` and `y`

data/com/mojang/blaze3d/systems/RenderPass.mapping
PiTheGuy: I've only ever used OpenGL names and I don't even see any references to Vulkan in the source code so I don't see why we should use their names instead.

data/com/mojang/blaze3d/buffers/Std140SizeCalculator.mapping
RogueLogix: Ahh, right, this class doesnt take params, it has different functions for each type, my bad.

data/com/mojang/blaze3d/opengl/BufferStorage.mapping
RogueLogix: _amazing, fair enough then_

data/com/mojang/blaze3d/systems/RenderPass.mapping
RogueLogix: Because VK's names for these parameters are consistent across the different draw commands, while GL isn't, and in some cases match what VK uses exactly.

VK's draw commands are
`vkCmdDraw(vertexCount, instanceCount, firstVertex, firstInstance)`
`vkCmdDrawIndexed(indexCount, instanceCount, firstIndex, vertexOffset, firstInstance)`
`vkCmdDrawIndirect()` which takes `VkDrawIndirectCommand` structs, using `vertexCount, instanceCount, firstVertex, firstInstance`
and `vkCmdDrawIndexedIndirect` which has a similar struct with `indexCount, instanceCount, firstIndex, vertexOffset, firstInstance`
the other draw commands not listed have consistently named parameters.
these are consistent within themselves, with only firstVertex vs vertexOffset (which have different meanings) being different.

GL on the other hand, is not consistent within itself.
`glDrawElementsInstance(mode [not present], count [indexCount], type [not present], indices [firstIndex], instanceCount)` exists along side
`glDrawElementsInstancedBaseVertexBaseInstance` which can be defined with different argument names
the ARB extension uses `mode, count, type, indicies, primcount, basevertex, baseinstance` while the core function uses (for the same meaning) `mode, count, type, indicies, instancecount, basevertex, baseinstance`
`glDrawElementsIndirect` and `glMultiDrawElementsIndirect` then take a struct with 
`count, instanceCount, firstIndex, baseVertex, baseInstance`
the `*drawArrays*` commands are similarly inconsistent in this

Picking which name is the correct one to use from the GL spec is a cluster, because these calls aggregate down to multiple of them internally. VK doesn't completely solve this (vertexOffset vs firstVertex for some internal `GlCommandEncoder` calls), but does for all of the user facing interface calls.

There is also substantial VK influence over the design of the API. As examples. GL has no concept of a RenderPass (but [VK does](https://registry.khronos.org/vulkan/specs/latest/man/html/VkRenderPass.html)), nor a concept of a CommandEncoder (but [VK does, CommandBuffer](https://registry.khronos.org/vulkan/specs/latest/man/html/VkCommandBuffer.html)), and GL also does not require use of texture views (but [VK does](https://registry.khronos.org/vulkan/specs/latest/man/html/VkDescriptorImageInfo.html)) so its not out of place to use VK (based) names or concepts instead of GL ones.

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
RogueLogix: fair enough

data/com/mojang/blaze3d/systems/RenderPass.mapping
RogueLogix: This is not resolved. Each draw has its own vertex buffer, and potentially index buffer too. This needs to be given a more specific name. `indexBuffer` or `defaultIndexBuffer` would both be appropriate here. Simply "buffer" is not descriptive.

data/com/mojang/blaze3d/systems/RenderPass.mapping
RogueLogix: This is not resolved. There are two different sets of uniforms during a drawMultiple call. This specifies only the ones that will be dynamically set per draw call through the UniformUploader and bypasses validation on them being set ahead of time. This should _not_ have all uniforms passed into it as just "uniforms" implies. this is "uniforms that you will set via UniformUploader ahead of at least the first draw but probably for all of them individually and its ok if they aren't set yet"

data/com/mojang/blaze3d/systems/RenderPass.mapping
PiTheGuy: @ParchmentMC/mapping-reviewers any input on this?

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
RogueLogix: Assuming that line is coming from inside `GlCommandEncoder#createRenderPass`, that `debugGroup` is a parchment added name, and likely also incorrect.


Because this never gets saved to or read from anything we have mojmap for, the two options are that Mojang was consistent with themselves everywhere using "label" or with the underlying GL API (in which case this is "message"). In either case "debugGroup" is incorrect.

data/com/mojang/blaze3d/opengl/DirectStateAccess.mapping
```diff
@@ -18,6 +18,37 @@ CLASS com/mojang/blaze3d/opengl/DirectStateAccess
...
+		ARG 1 buffer
+		ARG 2 data
+		ARG 3 usage
+	METHOD bufferStorage (IJI)V
+		ARG 1 buffer
+		ARG 2 size
+		ARG 4 usage
+	METHOD bufferStorage (ILjava/nio/ByteBuffer;I)V
+		ARG 1 buffer
+		ARG 2 byteBuffer
```
RogueLogix: missed this one for "data"

data/com/mojang/blaze3d/opengl/GlDebugLabel.mapping
```diff
@@ -16,3 +16,5 @@ CLASS com/mojang/blaze3d/opengl/GlDebugLabel
 		ARG 0 capabilities
 		ARG 1 renderDebugLabels
 		ARG 2 enabledExtensions
+	METHOD pushDebugGroup (Ljava/util/function/Supplier;)V
+		ARG 1 name
```
RogueLogix: related to other pushDebugGroup discussion, this is inconsistent

data/com/mojang/blaze3d/systems/CommandEncoder.mapping
RogueLogix: asked and got confirmation, this is "label"

data/com/mojang/blaze3d/systems/RenderPass.mapping
Lanse505: I'd personally say go with the more consistent naming scheme I.E. ``instanceCount`` since that is also a name used in OpenGL it's the more "generally used" name in this context.

RogueLogix: Notable, i am suggesting that _all_ parameters to _all_ draw calls should follow what VK calls them, not just "primCount" vs "instanceCount".

this would include replacing "index" with "indexCount" and for "draw" using "firstVertex" instead of "firstIndex" (which is outright wrong, though not changed by this PR) and "vertexCount" instead of "indexCount" (which is again, outright wrong, its not an indexed draw command). I dont think baseVertex/vertexOffset is a thing in any vanilla draw command, so that doesn't apply here.

Lanse505: As long as it doesn't go against Mapping Standards #3 I don't personally have an issue with it.
Though I'd like @ChampionAsh5357 's input here aswell.

ChampionAsh5357: It sounds reasonable to me, but given the change, I think there should be some basic javadocs for the parameters. I would rather not assume that people understand VK names. This, or at least the javadocs portion, could be a separate PR though.

PiTheGuy: As the person who does the majority of the porting, I feel I would take on most of the burden from such a change. I think it would be much easier for me to use the OpenGL names as those functions are directly referenced in the source code, as opposed to looking up VK functions and figure out which one applies.

Lanse505: In that case do what feels best for you in this case @PiTheGuy .
If we or @RogueLogix feel that names could be better suited using the VK names, a second PR can be opened at a later date improving the mappings. I think the priority here would be to get the mappings approved, merged and then we can improve upon them <3

RogueLogix: I'm ok with this being addressed in a separate PR.

Lanse505 (dismissed):
LGTM

data/com/mojang/blaze3d/systems/RenderPass.mapping
```diff
@@ -2,30 +2,61 @@ CLASS com/mojang/blaze3d/systems/RenderPass
...
 	METHOD drawMultipleIndexed (Ljava/util/Collection;Lcom/mojang/blaze3d/buffers/GpuBuffer;Lcom/mojang/blaze3d/vertex/VertexFormat$IndexType;)V
 		ARG 1 draws
 		ARG 2 buffer
 		ARG 3 indexType
+	METHOD drawMultipleIndexed (Ljava/util/Collection;Lcom/mojang/blaze3d/buffers/GpuBuffer;Lcom/mojang/blaze3d/vertex/VertexFormat$IndexType;Ljava/util/Collection;)V
+		ARG 1 draws
+		ARG 2 buffer
+		ARG 3 indexType
+		ARG 4 uniforms
+	METHOD drawMultipleIndexed (Ljava/util/Collection;Lcom/mojang/blaze3d/buffers/GpuBuffer;Lcom/mojang/blaze3d/vertex/VertexFormat$IndexType;Ljava/util/Collection;Ljava/lang/Object;)V
```
RogueLogix, line 13: I see there's two of these added, should there be two added?
I didn't catch this the first time, but theres only one `drawMultipleIndexed`. iirc the former definition was part of a snapshot. This is the one that will get used though, and its names are fine.

RogueLogix (approved):

data/com/mojang/blaze3d/systems/RenderPass.mapping
Lanse505: @sciwhiz12 ?

RogueLogix: looking at gitblame, seems the other one is from 25w16a, not really a problem though.

PiTheGuy: I would have thought `sanitizeData` would have removed the old one. It was my understanding that was why there was so many removals on these update PRs, because it removes all the clutter from previous snapshots.

Lanse505: Not a big problem, but pinging Sci because this should be getting fixed by our sanitize stuff I think 🤔

sciwhiz12: Found the problem -- I haven't been automatically generating Blackstone exports, so the sanitize data task was running without it, which means it doesn't do every sanitation task.

In fairness to me, it does remark that in the logs:
```
> Task :sanitizeData
Unable to resolve 'blackstone' configuration
Sanitizers in use: [non-existing members, bouncer data mover, enum valueOf remover, synthetics, descriptor parameter indexes]
No Blackstone metadata loaded, sanitation may not have any effects
Sanitizing mapping data in 'D:\Dev\Parchment\Mappings\data'
```

But it doesn't absolve me from not generating Blackstone exports for the past few MC versions. I'll be hooking it up (or rather, @Matyrobbrt will do it for me) to NeoForged's n8n instance so it runs and generates for each new Minecraft version.

For now, I've generated exports for 1.21.4, 1.21.5, 1.21.6, and 1.21.7-rc1. If you run the `sanitizeData` task again, it'll properly download Blackstone and run all the sanitations. 

(Note: local running gives me a `1383 files changed, 69 insertions(+), 24307 deletions(-)`, so uh 😓)

Lanse505: We'll hold off on merging then until we can sanitize.

Lanse505 (approved):

Lanse505:
PR is now merged @RogueLogix if you want to open that PR for VK names you base it upon this commit :)

PR #341, opened 2025-09-23
Fix PalettedContainer$Data#createOrReuseData param

jpenilla:
```java
    private PalettedContainer.Data<T> createOrReuseData(@Nullable PalettedContainer.Data<T> data, int bits) {
        Configuration configurationForBitCount = this.strategy.getConfigurationForBitCount(bits);
```


data/net/minecraft/world/level/chunk/PalettedContainer.mapping
```diff
@@ -37,7 +37,7 @@ CLASS net/minecraft/world/level/chunk/PalettedContainer
 		ARG 1 countConsumer
 	METHOD createOrReuseData (Lnet/minecraft/world/level/chunk/PalettedContainer$Data;I)Lnet/minecraft/world/level/chunk/PalettedContainer$Data;
 		ARG 1 data
-		ARG 2 id
+		ARG 2 bits
```
marchermans, line 40: Would `bitCount` not be a better option?

marchermans (changes requested):

PR #343, opened 2025-10-07
Update to 1.21.10

PiTheGuy:


Spinoscythe (approved):

RedstoneDubstep:
When updating our mod to 1.21.10, we discussed how to name the new boolean param that `entityInside` received, and had to figure out what that param even does. Ingame testing revealed that this parameter is always `true` except in the one tick when the entity has left the block (and `entityInside` is still called again for some reason). So we decided to name this parameter `stillInside` in our overrides. Your `intersects` definitely work too, but I thought I'd give some creative input in case you think `stillInside` (or similar) is more descriptive or something.

Matyrobbrt (approved):
I think intersects is fine, if the parameter creates confusion we can revisit it later.

PR #345, opened 2025-12-10
Update to 1.21.11

PiTheGuy:
This is it. The final obfuscated version of Minecraft. This is the culmination of the hundreds if not thousands of hours I've put into mapping parameters over the last 4.5 years.


Matyrobbrt (approved):

Matyrobbrt:
Unless a hotfix is needed for 1.21.11 and with 26.1 snapshots already starting, this is likely to be one of the last if not the last versions of Parchment. Thank you for your contributions over the years, we all appreciate it.

